PR #19, opened 2015-01-18
Add  redirect_uri to the OAuth2 Access Token Request

0111b:
According to [RFC](http://tools.ietf.org/html/rfc6749#section-4.1.3) redirect_uri is required, if the "redirect_uri" parameter was included in the authorization request, and their values must be identical


dongri:
Thank you!

PR #28, opened 2015-03-03
Support for Swift 1.2

rvangraan:
- Changed Byte type to UInt8
- Changed visibility of methods in OAuthSwiftClient to public to support inheriting/overiding
  when the library is used with "use_frameworks!" in Cocoa Pods


dongri:
Thank you!

PR #32, opened 2015-03-08
Oauthswift salesforce

quintonwall:
I added support for Salesforce authentication. It is all pretty straight forward except I added a call to remove encoding on code response keys. Salesforce definitely sends keys that have special characters which require URLEncoding (and were breaking my app until I removed them). I suspect other providers may have similar issues


dongri:
Thank you!

PR #36, opened 2015-03-19
Allow optional skipping the oauth verifier

gburgyan:
Optionally (defaulting to the spec-correct function) allow the skipping of the oauth_verifier to deal with non-spec-compliant oauth providers. SmugMug is such a provider. Some references to this can be found at http://www.dgrin.com/archive/index.php/t-212289.html (dgrin.com is smugmug's forum)

Thanks,
-George


dongri:
Thank you!

PR #37, opened 2015-03-20
Allow base authorizeUrl to contain query string

danielhill:
Currently, _OAuth1Swift_ and _OAuth2Swift_ append query string parameters to _authorizeUrl_ with no regard as to whether a query string already exists, resulting in URL's like http://example.com/oauth/authorize?display=touch?oauth_token=12345

This pull request changes this functionality to check to see if a _?_ is already present in the URL. If so, it will append a _&_ before further parameters, otherwise a _?_ is used.

I've also updated the FitBit demo to pass in the display=touch parameter, as making that work was the whole point of this modification.


dongri:
Thank you!

Issue #31, opened 2015-03-08
Serious compile issues


TarangKhanna:
The error is: unknown>:0: error: could not build Objective-C module 'OAuthSwift'
It works and suddenly it gives this error on build. I have tried this on 3 different laptops and the solution to clean build and delete derived data does not work all the time. Sometimes it shows one error(aforementioned) or it shows 21 errors
We spent over 4 hours with apple engineers on this in a hackathon with no progress. 
Thank you for your work!

danshev:
I believe related:

When I open the Demo in the newest version of xCode 6.2 (6C131e), I also get a lot of compile errors.  Many are in the `OAuth1` and `OAuth2` due to the use of `as!`.  When I remove the exclamation, the errors go away, except for a few:

`Cannot downcast from 'UIViewController?' to non-@objc protocol type 'WebViewProtocol'`

on the line:

    if let webView = self.webViewController as WebViewProtocol {

Chrisell:
Same issues here. Xcode Version 6.2 (6C131e) as well.

ramki1979:
How to build this with 6.1.1, I am getting the following errors for length & size methods in String+OAuthSwift.swift

```swift
'String' is not convertible to 'Range<I>'
```

gabriel-jones:
There are a few things which Dongri should change for this project, but I'll tell you guys anyway so you know what to fix. There are a few main types of errors. First, the down casting he does looks like this:
```  as! String ```
Which will make an error. What you have to do is write this:
```  as String! ```

--------------------------------------------------------------------------------------------

Next error is what danshev said: 
``` if let webView = self.webViewController as? WebViewProtocol  ```
Has an error. Easiest way to fix this is to go to ```OAuthWebViewController.swift```  and find this piece of code on line 13: 
` protocol WebViewProtocol `
and change it to: 
` @objc protocol WebViewProtocol`

--------------------------------------------------------------------------------------------

Next error is what ramki1979 said: there are some string errors in some files. One you can fix is this:
` if count(queryString.utf16) == 0 `
Just change it to a property, not a function: 
`if queryString.length() == 0 `


Bascially, if you ever see 
`count(stringName.utf16)`
Just change it to 
`stringName.count`


Also another example under `String+OAuthSwift.swift` you will find this: 
`    func length()->Int{
        return count(self)
    }`

which should just return: 
`return self.count()`
--------------------------------------------------------------------------------------------
Hope I helped you guys. Tell me if I missed anything!

[...]

PR #46, opened 2015-04-09
Minor fixes to build in 6.2

Automatt:
Implements fixes identified by @gabriel-jones in #31 to build in Xcode 6.2


gabriel-jones:
Another fix: my bad, do not change to ```stringName.length()``` but to ```stringName.utf16count```

PR #56, opened 2015-04-29
fix code extraction for webview

corinnekrych:
When working with WebView test the url fragment is not nil AND is not an empty string. Since Swift 1.2 a cast is done in String that can now contains empty string.


dongri:
Thank you

PR #62, opened 2015-05-11
Add OSX, better way to present custom web controller, fix one bug

phimage:
1 / Fix OAuth1 if callback url returned by service has no query (only fragment)

2 / Add compatibility with OSX, add demo project inspired from iOS one (many code to factorise)

3/ Normally a controller is not responsible to choose its parent controller as you do with the code 
```swift
UIApplication.sharedApplication().keyWindow?.rootViewController?.presentViewController(
                        self.webViewController!, animated: true, completion: nil)
```
A parent controller, not only the root one, could do it (moreover this root controller doesn't exist in OSX)
Maybe the web view controller could be in a storyboard, a segue must be performed to show it (with maybe a custom animation)
Maybe a web view is already present in the application, and no web view must be shown, only and url update of the current one

That's why 
- I rename the protocol WebViewProtocol to OAuthSwiftURLHandlerType, move to its own file,
- Replace in OAuth objects the property webViewController by an authorize_url_handler: OAuthSwiftURLHandlerType
- Make a default implementation to open URL externally has you do if there is no webViewController assigned to the OAuth object

In consequence, no more UIKit(or AppKit for OSX) reference to the core framework
Users of framework can make their own code to show the web view which handle the url
OAuthWebViewController can always be extended with default behaviour (this object could be excluded from core framework and be part of a submodule in podspec)

### DIFF
--- a/OAuthSwift.podspec
+++ b/OAuthSwift.podspec
@@ -2,13 +2,14 @@ Pod::Spec.new do |s|
   s.name = 'OAuthSwift'
   s.version = '0.3.4'
   s.license = 'MIT'
-  s.summary = 'Swift based OAuth library for iOS.'
+  s.summary = 'Swift based OAuth library for iOS and OSX.'
   s.homepage = 'https://github.com/dongri/OAuthSwift'
   s.social_media_url = 'http://twitter.com/dongriat'
   s.authors = { 'Dongri Jin' => 'dongriat@gmail.com' }
   s.source = { :git => 'https://github.com/dongri/OAuthSwift.git', :tag => s.version }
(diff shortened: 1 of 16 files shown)


hopye:
This would be nice, i really need to make the web view under my navigation controller

hopye:
in the mean time, someone knows how to do it manually right now ? im kinda of a newbi..

dongri:
Hi @phimage 

Thank you!

https://github.com/dongri/OAuthSwift/commit/43a05b2c7c59b01a12e0cfe0a688cc64e6cfd931

phimage:
you are welcome
sorry for the "key" instead of "oauth_token" (just testing one website api...)

PR #72, opened 2015-05-31
Pass along response parameters to the caller

munkius:
I was integrating an OAuth app with mine and spotted they used a so-called `refresh_token`, which was pass along with the response parameters. I made a change to pass back all the response parameters, as some apps also send other (custom) data to the caller.


alexdd55:
:+1:

dongri:
@munkius Thank you!

Issue #74, opened 2015-06-06
Yelp example


derrickshowers:
Is there any way to add an example for how this can be used for Yelp? Seems that [their API](https://www.yelp.com/developers/documentation/v2/authentication) is different from others because the token is already provided and therefore there is not an endpoint available to retrieve the token. I'm hoping there is a way with your framework to manually provide the token?

derrickshowers:
Submitted a [pull request](https://github.com/dongri/OAuthSwift/pull/75) for this.

phimage:
I will close this issue
I repost here my comment on PR

Instanciate the OAuthSwift object as usual
Then access, using getter , the client and the credential, then set the token & token secret from conf using property setter

```swift

 let oauth = OAuth1Swift (
...
 )

oauth.client.credential.oauth_token =  mytoken
oauth.client.credential.oauth_token_secret = mytoken_secret
oauth.client.credential.oauth_header_type = "oauth1" // just a fix because not initialized at init yet

```

PR #75, opened 2015-06-06
added yelp to demo

derrickshowers:
Nice addition to the demo since Yelp's OAuth works a bit differently from the rest. The token and token secret values are not dynamic, provided on a per developer basis. This skips right to using the `OAuthSwiftClient` class, setting all four values.

### DIFF
--- a/OAuthSwiftDemo/Constants.swift
+++ b/OAuthSwiftDemo/Constants.swift
@@ -98,4 +98,10 @@ let Tumblr =
     "consumerKey": "***",
     "consumerSecret": "***"
 ]
-
+let Yelp =
+[
+    "consumerKey": "***",
+    "consumerSecret": "***",
+    "accessToken": "***",
+    "accessTokenSecret": "***"
+]
--- a/OAuthSwiftDemo/ViewController.swift
+++ b/OAuthSwiftDemo/ViewController.swift
@@ -11,7 +11,7 @@ import OAuthSwift
 
 class ViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
 
-    var services = ["Twitter", "Flickr", "Github", "Instagram", "Foursquare", "Fitbit", "Withings", "Linkedin", "Linkedin2", "Dropbox", "Dribbble", "Salesforce", "BitBucket", "GoogleDrive", "Smugmug", "Intuit", "Zaim", "Tumblr"]
+    var services = ["Twitter", "Flickr", "Github", "Instagram", "Foursquare", "Fitbit", "Withings", "Linkedin", "Linkedin2", "Dropbox", "Dribbble", "Salesforce", "BitBucket", "GoogleDrive", "Smugmug", "Intuit", "Zaim", "Tumblr", "Yelp"]
     
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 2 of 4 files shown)


phimage:
Could you add it for OSX demo too?

and add message to close #74 : https://help.github.com/articles/closing-issues-via-commit-messages/
only if you think there is no more to discuss in the issue (maybe not, we can talk about the issue in general, and solution could be provide a way to customize parameters)

derrickshowers:
This fixes #74 

Added for OSX demo. I'd imagine whoever uses this in an actual app would have some way of dynamically setting parameters, so it would seem appropriate to leave them in the example as is. Let me know if you were thinking of something different.

derrickshowers:
@phimage Can this pull request be merged? I see there is conflicts now, and would be happy to resolve, but first wanted to check if you are planning to merge. Thanks!

phimage:
@derrickshowers I am not responsible for this project, my PR is not merged too

derrickshowers:
Oh! My bad, I just assumed. :smile:

sephethus:
How do I use this OAuthSwiftClient when making requests to search yelp? I'm getting an error that says cannot find initializer for type OAuthSwiftClient that accepts an argument of type (consumer key: string... yada yada yada.

sephethus:
I'm using it this way:

let oauthYelp = OAuthSwiftClient(consumerKey: CONSUMER_KEY, consumerSecret: CONSUMER_SECRET, accessToken: TOKEN, accessTokenSecret: TOKEN_SECRET)
        let params: [String: String] = [ "location": "Philadelphia", "term": "seafood"]
        let toEscape = YELPAPI_HOST + YELPSEARCH_PATH
        let urlString = toEscape.stringByAddingPercentEscapesUsingEncoding(NSUTF8StringEncoding)!
        oauthYelp.get(urlString,
            parameters: params,
            success: { (data, response) -> Void in
                let json: NSDictionary = NSJSONSerialization.JSONObjectWithData(data, options: nil, error: nil) as! NSDictionary
                println(json)
            }) { (error) -> Void in
                    println("there was an error: \(error)")

        }

sephethus:
I got it working! Thanks so much for this!!!!!!

tagatac:
Merge this! It's useful for OAuth n00bs :)

phimage:
Before merging, I think there is a more conventionnal way to do it.
Instanciate the OAuthSwift object as usal
Then access, using getter , the client and the credential, then set the token & token secret from conf using property setter

Its seems weird that token are not dynamic, not really oauth

tagatac:
Can you elaborate why this is better than constructing a client directly?

phimage:
Because this the generic way to do it when you have already got the tokens (could be in a previous app launch and you have stored it)

Then the client class could become a protocol, allowing to replace by other network api or init could become private (I don't think there is a reason to be public except doing unit test)

tagatac:
Okay, thanks for the explanation. Also note that it's necessary to manually set the credential's oauth_header_type="oauth1". Normally this is done when the OAuth object requests a token, but this never happens for the Yelp API. This PR doesn't do that; maybe this is new logic in OAuthSwift since the PR was opened.

phimage:
Yes I know for the version, its new, before it was a boolean (default false means oauth1), that's why I make PR #121 to set version at init and put into client oauth1 as default

phimage:
I will close due to conflict, you can make a new one or reopen if you use code from #74

PR #78, opened 2015-06-26
Allow to modify client attributes Fix #26

### LINKED ISSUE #26: Keep user logged
Hi,
suppose that I have an app with a Twitter login button. After the login, I'm given back the credentials and I can store them in NSUserDefaults.standardUserDefaults()
If the user runs the app again, and I see that he has auth credentials already set, how can I set up the oauth client so that it uses them without requiring the login again?

Also, once the user it logged in, how to get info of the logged user, such is user id etc?

Thanks

phimage:
to set a previously computed token
```swift        
oauthswift.client.credential.oauth_token = XXX
oauthswift.client.credential.oauth_token_secret = XXX
```

See #26, allow to set value already computed with previous session

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -18,7 +18,7 @@ public class OAuthSwiftClient {
         static let signatureMethod = "HMAC-SHA1"
     }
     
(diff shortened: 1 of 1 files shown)


dongri:
Hi @phimage 

Thank you!

PR #79, opened 2015-06-29
Passing along response parameters to the caller also for OAuth1

phimage:
Inspired from 647a3ab5a4d461acd9ab7620a9fad5c34c44eb1a (add parameters for OAuth2 into TokenSuccessHandler) and to keep the coherence, I have hadded `parameters` to `TokenSuccessHandler` for OAuth1


phimage:
conflict resolved

this will fix #81

PR #83, opened 2015-07-09
Handle cancellation of the Authentication

Poupouch:
Using LinkedIn, but working for Facebook and Google as well, I want to know if my user cancelled the authentication.

screen shot 2015-07-09 at 1 47 54 pm


dongri:
Thank you! @Poupouch

PR #98, opened 2015-08-26
Add Uber into OSX demo app

phimage:
If you want I can work on factorize codes between the two demo app, to just have to add only in one place


dongri:
@phimage Thank you!

PR #108, opened 2015-09-21
Fix error: `'parentViewController' is only available on OS X 10.10 or…

s-aska:
I'm set deployment_target 10.10 :smiley:


phimage:
We can do better with swift 2 to keep compatibility with 10.9...
and `OAuthWebViewController` is not a necessary class, just an helper for lazy people

```swift
if #available(OSX 10.10, *) {

} 
```

PR #109, opened 2015-09-22
OS X BuildableName

s-aska:
BuildableName even OS X should be the OAuthSwift, because cocoapods will always generate a OAuthSwift.framework.

### DIFF
--- a/OAuthSwift.xcodeproj/project.pbxproj
+++ b/OAuthSwift.xcodeproj/project.pbxproj
@@ -100,7 +100,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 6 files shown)


phimage:
xcode target and cocoapods are two things differents
You can have multiple target with many different names, with specific chosen files likesubspec in cocoapod

If you use cocoapod yes you do import OauthSwift, cocoapod want just an unique name, but cocoapod do not drive the xcode use

s-aska:
I'm writing a module that depends on OAuthSwift.
It is desired to correspond to cocoapods and Carthage.
But Carthage would generate a OAuthSwiftOSX.framework. : /
I want to try if there is writing podspec for generating the OAuthSwiftOSX.framework to cocoapods.
Because I want to unify the BuildableName in SwiftyJSON was it to the reference. :sweat:
And sorry if there is a deficiency in English.

phimage:
I am french so my english is also bad, like my japanese 

Yes i am agree if you use Carthage,  which use target

PR #111, opened 2015-09-24
FIX #110 merge query and fragment parameters instead of choosing one

### LINKED ISSUE #110: Foursquare authentication failing
Hi, I'm using OAuthSwift library on a project, and while doing Foursquare authentication I found something interesting inside the file OAuth2Swift.swift, when calling authorizeWithCallbackURL function I see that the returned url has both a query string and a fragment, it will be something like.

https://somehost/somepath?access_token=${atoken}#_=_

and because of that the responseParameters lose the access_token value with the code:

```
            var responseParameters: Dictionary<String, String> = Dictionary()
            if let query = url.query {
                responseParameters = query.parametersFromQueryString()
            }
            if ((url.fragment) != nil && url.fragment!.isEmpty == false) {
                responseParameters = url.fragment!.parametersFromQueryString()
            }
```

Cheers
Ignacio

phimage:
in oauth1 the `else`is removed like in oauth2

and parameters dictionary are merged for the two versions

see my comment in issue #110, there is another fix

### DIFF
--- a/OAuthSwift/OAuth1Swift.swift
+++ b/OAuthSwift/OAuth1Swift.swift
@@ -61,9 +61,10 @@ public class OAuth1Swift: NSObject {
                 let url = notification.userInfo![CallbackNotification.optionsURLKey] as! NSURL
                 var parameters: Dictionary<String, String> = Dictionary()
                 if ((url.query) != nil){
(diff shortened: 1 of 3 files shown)


dongri:
@phimage Thank you!

PR #113, opened 2015-09-27
Move dico extensions from Util to good file

phimage:
In my previous PR I put dictionary merge function into Util.swift
I don't see extensions files into collapsed group Extensions

So this commit just move into the good place and add operator + that use dictionary join function


dongri:
@phimage Merci!

phimage:
どう致しまして

PR #114, opened 2015-09-27
Allow to use OAuthSwift with custom network call or other network API

phimage:
- Factorising method to create `OAuthSwiftHTTPRequest`
- Move method about credential into `OAuthSwiftCredential` instead of `OAuthSwiftClient`
- Extract method to create headers from credential
- Add method to create `NSMutableURLRequest` from `OAuthSwiftHTTPRequest` (just calling the static one)

I have just extracting and moving codes, no change

With this public methods its now easy to create a custom `NSMutableURLRequest` with headers from credential, or just get the one created by this api

We can for instance use Alamofire like that
```swift
 extension OAuthSwiftHTTPRequest : URLRequestConvertible {

public var URLRequest: NSMutableURLRequest {
        return try! self.makeRequest()
}
```
or just by adding headers as parameter of `request` Alamofire function using `makeHeaders`new function from `OAuthSwiftCredential`

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -12,12 +12,7 @@ import Accounts
 var dataEncoding: NSStringEncoding = NSUTF8StringEncoding
 
 public class OAuthSwiftClient {
-    
-    struct OAuth {
-        static let version = "1.0"
-        static let signatureMethod = "HMAC-SHA1"
-    }
-    
+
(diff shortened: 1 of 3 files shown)


phimage:
Fix conflict with c6cea89e66c101dccd20dced608b7f9625d36c36 (I have done the same thing on makeRequest and add one function near that, that's why conflict existed)

dongri:
@phimage Thanks!

PR #123, opened 2015-10-07
Check the http url response's status code before using it

sdduursma:
Previously the NSHTTPURLResponse's status code passed to the completion handler of `dataTaskWithRequest` in `start()` in OAuthSwiftHTTPRequest was only checked after the response and data objects had been used. This PR fixes that.


benjaminboxler:
Close this PR

PR #124, opened 2015-10-08
Don't let the app crash when there is no access_token in oauth2 response.

srna:
Don't let the app crash when there is no access_token in oauth2 response.


phimage:
:+1:

PR #125, opened 2015-10-09
Fix crash where there is no data in the response.

erndev:
check for valid `data`,  and if it's null, return calling the failure closure.


iiiyu:
:+1:

phimage:
I close this one as duplicate of more complet #132

PR #132, opened 2015-10-26
Added protections on network timeout.. force unwrapping of 'data' was…

mikeMTOL:
… causing a unwrap of nil object crash..
blocks should not refer to 'self' directly and must use an unowned reference to self (memory leaks happen if not)


phimage:
:+1:

Issue #140, opened 2015-11-09
OAuth1 exchange fails due token secret being URL encoded.


pculligan:
In using the OAuth1 provider via the OAuthSwiftDemo the initial token request works, but once the credential's `oauth_token_secret` is assigned, all signatures fail.

Step 1: Request token
``` 
POST /platform/oauth/request_token HTTP/1.1
Authorization: OAuth realm="", oauth_callback="oauth-swift%3A%2F%2Foauth-callback%2Fjdcloud", oauth_consumer_key="johndeere-3WuO2CyXvjOAvq5HyQ4C657B", oauth_nonce="1446926761", oauth_signature="ej%2FrsK%2BYoKzQJr8vXqRLc3X81I0%3D", oauth_signature_method="HMAC-SHA1", oauth_timestamp="492373033280972651", oauth_version="1.0"
```

Note that there is no token at this stage, so the signature is fine, and we get a token and secret back.

```
HTTP/1.1 200 OK
oauth_token=cb88aa77-6e26-499a-806a-52590d57785b&oauth_token_secret=Q3StNc9bb4n4AF%2FvyiIeiRjwmjqsaBxWWF73wUpqxjjYADLuXRsPrE2sgqLedQmdABSjL9SqDGKEwrc5ZXdow1LiNk7kWQ8K2OY6xp2bRJI%3D&oauth_callback_confirmed=true
```

When we try to authorize during step 2 however, I get a 401, constantly. After tracing, I found that in the OAuth1Swift.swift file the method takes the callback URL directly.

```Swift
// 1. Request token
public func postOAuthRequestTokenWithCallbackURL(callbackURL: NSURL, success: TokenSuccessHandler, failure: FailureHandler?) {

...

}

// 3. Get Access token
func postOAuthAccessTokenWithRequestToken(success: TokenSuccessHandler, failure: FailureHandler?) {

...

}
```

Within the method, the QueryString is parsed by the `parametersFromQueryString` method into `parameters`. 

```
    func parametersFromQueryString() -> Dictionary<String, String> {
        var parameters = Dictionary<String, String>()

        let scanner = NSScanner(string: self)

        var key: NSString?
        var value: NSString?

        while !scanner.atEnd {
            key = nil
            scanner.scanUpToString("=", intoString: &key)
            scanner.scanString("=", intoString: nil)

            value = nil
            scanner.scanUpToString("&", intoString: &value)
            scanner.scanString("&", intoString: nil)

[...]

Issue #80, opened 2015-07-06
POST request Authentication Failed


gabriel-jones:
I am trying to upload a piece of data to Intuit's Quickbooks API, by doing something like this:

```
//Set parameters for request
        let parameters: [String: AnyObject] = [
            "NameOf":"Employee",
            "EmployeeRef":[
                "value":"64"
            ],
            "CustomerRef":[
                "value":"21"
            ],
            "ItemRef":[
                "value":"1"
            ],
            "BreakHours": 1,
            "BreakMinutes":0,
            "StartTime": "example start time",
            "EndTime": "example end time",
            "Description": "example description",
            "domain": "QBO",
            "sparse": false
        ]
        //Send a POST request to the Quickbooks API endpoint
        oauthswift.client.post("https://sandbox-quickbooks.api.intuit.com/v3/company/<companyID>/timeactivity", parameters:  parameters, success: {
            data, response in
            println("success")
            }, failure: {
            error in
            println(error.localizedDescription)
        })

```

But I get back an error from the server saying the content type is not valid. So I tried changing (in the OAuthSwiftClient) where it says `request.encodeParameters = true` to `request.encodeParameters = false`, and this gives me the error from the server saying the application's authentication failed, and the OAuth signature is not included in the response string.

I should probably add that doing a GET request works fine.

Has anybody else had these problems?

phimage:
You set `request.encodeParameters = false` in framework code to force encode json (and add json header), code maybe not reachable without hacking the framework...

For OAuth v1 there is a method `OAuthSwiftClient.authorizationHeaderForMethod`
I think your json is merged into the Authorization header, so the website does'nt understand anything


You can test by modifying other code into `OAuthSwiftClient`
```swift
 func request(url: String, method: String, parameters: Dictionary<String, AnyObject>, success: OAuthSwiftHTTPRequest.SuccessHandler?, failure: OAuthSwiftHTTPRequest.FailureHandler?) {
```
replace `parameters` here with empty dictionnary
```swift
request.headers = ["Authorization": OAuthSwiftClient.authorizationHeaderForMethod(method, url: url, parameters: parameters, credential: self.credential)]
 ```

gabriel-jones:
@phimage Could you give an example of all changes to the request function in OAuthSwiftClient?

phimage:
I am just taking about one little change already described upper with function name and line of code
empty dico :  Dictionary<String, AnyObject>()
instead of parameters

[...]

PR #141, opened 2015-11-09
Closes Issue #140 and #142.

pculligan:
Added decoding the URL-encoded `oauth_token_secret` to the OAuth1 flow (in Step 1 and Step 3) so that it would complete. Previous escaped values were preventing proper OAuth signatures.

### DIFF
--- a/OAuthSwift/OAuth1Swift.swift
+++ b/OAuthSwift/OAuth1Swift.swift
@@ -103,8 +103,12 @@ public class OAuth1Swift: NSObject {
             data, response in
             let responseString = NSString(data: data, encoding: NSUTF8StringEncoding) as String!
             let parameters = responseString.parametersFromQueryString()
-            self.client.credential.oauth_token = parameters["oauth_token"]!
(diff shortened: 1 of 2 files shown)


phimage:
Thanks for the PR
There is now conflict on makeRequest, can you merge dongri master into your own master to resolve the conflict
`method`(String) become `method.rawValue`(enum)

pculligan:
Changes merged, committed, pushed. New to GitHub flow - let me know if I've missed anything.

phimage:
Just in case, I remove query string using `NSUrlComponent`and remove some unwrap `!` but not tested (I will when I can unit tests)
fdfb368ed2357354a0a16b347e03c829647bbb22

phimage:
Previous comment tested, Unit test added

I revert some change here due to my comprehension of the code after study #80
`OAuthSwiftHTTPRequest` take the previous value, not the combined ones because when doing a POST, `parameters` is not in url query string but in HTTP Body

So the is also a bug in signature when doing a `POST`, because parameters must not be into signature calculation as I understand, and more bugs on the internal use of the post function(if no additional parameters, no body encoded...)...

PR #144, opened 2015-11-09
Fix #137 update readme with 'parameters' into callback

### LINKED ISSUE #137: Sample code in the readme is incorrect for oauth2
It looks like the signature for the success callback has changed -- and should include 'parameters'.

The current code does not compile.

The fix is below

```
oauthswift.authorizeWithCallbackURL( NSURL(string: "oauth-swift://oauth-callback/instagram"), scope: "likes+comments", state:"INSTAGRAM", success: {
    credential, response, **parameters** in
    println(credential.oauth_token)
}, failure: failureHandler)
```

phimage:
Just edited the readme to add missing `parameters`
as requested here #137


phimage:
will edit readme online

Issue #81, opened 2015-07-08
Get login info from Callback?


gabriel-jones:
How could I get the employee's unique ID from them logging in to the app connect page in Quickbooks? In other words, how can I access the callback's parameters sent back by Quickbooks so I can use that for other tasks?

phimage:
Provide some codes, api link if it is not "intuit" api. Is it OAuth1?

gabriel-jones:
The API uses OAuth1. It is the Quickbooks API, and calls are made to this endpoint: "https://sandbox-quickbooks.api.intuit.com/v3/company/<company_id>/", and the login is at this link: https://appcenter.intuit.com/Connect/Begin

phimage:
maybe PR #79 is the solution, parameters returned by server is added to success callback

phimage:
I closed my PR #79 due to conflit, need to make a new one

PR #152, opened 2015-11-29
Add parameters in OAuth1 callback

quangbuule:
This solves #81.
For instance: Twitter auth callback contains user_id that can't be found in any other APIs.


phimage:
Thank you
I can develop from now but by just watching, I think demo code must be updated too by adding parameters into callback before "in"
https://github.com/dongri/OAuthSwift/blob/master/OAuthSwiftDemo/ViewController.swift

And If you can readme.md too by adding parameters  also to oauth1 example

quangbuule:
Hi @phimage, I just updated the pull request. Please review again, thank you.

phimage:
missing also in test, I do it

PR #153, opened 2015-12-01
SafariURLHandler: missing "public" access for externally configurable variables

itchingpixels:
Hi,

Thanks for the great work of making this happen - it helped us not to reinvent the wheel big time!

I came across the issue where the following variables on SafariURLHandler weren't accessible from outside of the framework (missing public annotiation):

        var animated: Bool = true
        var delegate: SFSafariViewControllerDelegate?
        var presentCompletion: (() -> Void)?
        var dismissCompletion: (() -> Void)?

Since I needed to customise the behaviour of this class I needed to set these - and at the same time I thought these were probably intended to be externally configurable originally. (otherwise these are not used inside the framework, so there's no point of their existence:)

Cheers,

Mark

### DIFF
--- a/OAuthSwift/OAuthSwiftURLHandlerType.swift
+++ b/OAuthSwift/OAuthSwiftURLHandlerType.swift
@@ -53,15 +53,15 @@ import SafariServices
         var observers = [String: AnyObject]()
 
         // configure
-        var animated: Bool = true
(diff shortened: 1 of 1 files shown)


phimage:
Yes public is missing, on "factory" too
thanks for the PR

PR #158, opened 2015-12-14
Updated print and NSURL in readme

mattfxyz:
Updated `println` to `print` since it was replaced in the new versions of Swift 

NSURL in the example needs unwrapped.


phimage:
Thanks
One more println in OAuth2.0 example and we are good

mattfxyz:
Added it!

phimage:
I talk about `println`, not NSURL (`handle(url: NSURL!)` is wrong I think)

If you are able to stash your commit into one, or remove the last one is better to have a clean history

mattfxyz:
Alright. Reverted and recommitted the changes.

PR #159, opened 2015-12-16
Update example for fitbit OAuth2 response type

patoroco:
Hi,

Running example with `code` config retrieves an error:

`The operation couldn’t be completed. Could not get Access Token``

We need to update response type value to `token` to allow to retrieve this info correctly.

### DIFF
--- a/OAuthSwiftDemo/ViewController.swift
+++ b/OAuthSwiftDemo/ViewController.swift
@@ -271,7 +271,7 @@ extension ViewController {
             consumerSecret: serviceParameters["consumerSecret"]!,
             authorizeUrl:   "https://www.fitbit.com/oauth2/authorize",
             accessTokenUrl: "https://api.fitbit.com/oauth2/token",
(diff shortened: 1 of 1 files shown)


phimage:
I have tested it when I add it to the demo( with specific headers)
They change the flow?

The documentation talk always about "code" 
https://dev.fitbit.com/docs/oauth2/

patoroco:
Yes, it was strange for me too, but I was trying with `code` And it returned an error.

Following the documentation it looks like the correct value, so maybe I'm doing something wrong.

Please update me if you can test it.

Thanks for your quick reply.

phimage:
I make a test with current code and there is a parsing issue
In code JSON dictionary is casted to `String:String` but fitbit return a number in one of the parameter(expire) and dictionary parsed become `String:AnyObject`

I will fix the issue, so I close this PR but thank you to make me test ;)

patoroco:
Nice! 👌🏻

PR #160, opened 2015-12-18
add multiparts parameter to postMultiPartRequest

tkawaji:
Hi, thank you for great library.

I want to upload image with postMultiPartRequest, but I can't.
So I created OAuthSwiftMultipartData class for passing multiparat data,
and add method which can receive array of this instance.

How to use:
```
let fileData = OAuthSwiftMultipartData(name: "photo", data: imageData, fileName: "file.jpg", mimeType: "image/jpeg")
let multiparts = [fileData]
oauthswift.client.postMultiPartRequest("upload.json", method: .POST, parameters: parameters, multiparts: multiparts, success: {}, failure: {})
```

Thank you.

### DIFF
--- a/OAuthSwift.xcodeproj/project.pbxproj
+++ b/OAuthSwift.xcodeproj/project.pbxproj
@@ -86,6 +86,10 @@
 		C4E038A91BF4C9D9007CECEA /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = C4E038A81BF4C9D9007CECEA /* Constants.swift */; };
 		C4E038AA1BF4CACB007CECEA /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = C4E038A81BF4C9D9007CECEA /* Constants.swift */; };
 		C4FAE4361C061660000CE669 /* SignTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C4FAE4351C061660000CE669 /* SignTests.swift */; };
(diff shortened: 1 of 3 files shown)


phimage:
Thanks for the PR

Did you add your files to the xcode project, I see any change on it. Maybe not commited
(must be added into and checked for each OS, not test)

 Then could you change `multiparts: Array<OAuthSwiftMultipartData>` by  `multiparts: Array<OAuthSwiftMultipartData> = []` to not change method signature and allow framework user to not specify this parameters when not needed

Optionally `OAuthSwiftMultipartData`could be a struct

tkawaji:
Thank you for your reply.

I include this library with cocoapods.
In my Podfile,
```
pod "OAuthSwift", path: {WORKING_DIRECTORY}
```
Then I edit and add files in XCode, there is no other changed file in my working directory.
So I don't know which files should be commited.
And I couldn't test my code for only iOS.
Please teach me what to do.

FIles you pointed out has been changed and commited.

Thank you.

phimage:
OAuthSwift.xcodeproj/project.pbxproj

in diff you must see 10 lines changed with `OAuthSwiftMultipartData.swift`in it
if all os checked in xcode

tkawaji:
Thank you for advice.

I re-add OAuthSwiftMultipartData.swift to each target in XCode,
And I checked that there is no error.

phimage:
I will make some clean to `multiDataFromObject`method and I think move a part of your code into `OAuthSwiftMultipartData` file, if you search it after my change

tkawaji:
Thank you!
I will check it.

phimage:
38d77ac8d9a379206fb165cddb820e8941c8d32f
I extract an unique code to append multipart data to an NSMutableData
and use your `OAuthSwiftMultipartData` for the other case

I don't know what really do the other case, your solution with an object is better

Issue #156, opened 2015-12-13
Update twitter status,authorization problem


fewspider:
OAuthSwift version is `0.4.8`
Twitter API is `https://api.twitter.com/1.1/statuses/update.json`, 
you can visit `https://dev.twitter.com/rest/reference/post/statuses/update` for more detail.

When my parameters has `in_reply_to_status_id`,it return like below
```json
{
  "errors" : [
    {
      "message" : "Could not authenticate you.",
      "code" : 32
    }
  ]
}
```
But only send `status ` is ok, it is pretty odd.
My project use `https://github.com/Alamofire/Alamofire` to request the Twitter REST API.
I write a method to fetch header,do not forget replace the oauth_token,oauth_token_secret and Config,for some personal reason,i don't want to expose it:
```swift
private static func buildHeaders(requestMethod: OAuthSwiftHTTPRequest.Method, url: NSURL, parameters: [String: AnyObject]) -> [String : String] {
        let oauth_token = "" //replace it,for some personal reason,i don't want to expose it
        let oauth_token_secret = "" //replace it,for some personal reason,i don't want to expose it

        let c = OAuthSwiftClient(consumerKey: Config.consumerKey, consumerSecret: Config.consumerSecret, accessToken: oauth_token, accessTokenSecret: oauth_token_secret)
        let credential = c.credential

        let authorization = credential.authorizationHeaderForMethod(requestMethod,
            url: url,
            parameters: parameters)

        let headers = ["Authorization": authorization]
        return headers
    }
```
and set the header
```swift
private static func setHeaders(mutableURLRequest: NSMutableURLRequest, oauthMethod:OAuthSwiftHTTPRequest.Method, url: NSURL, parameters: [String: AnyObject]) -> NSMutableURLRequest {
        let headers = Router.buildHeaders(oauthMethod, url:url, parameters: parameters)
        for (field, value) in headers {
            mutableURLRequest.setValue(value, forHTTPHeaderField: field)
        }
        return mutableURLRequest
    }
```
The class is:
```swift
//
//  Router.swift
//  Twicrow
//
//  Created by fewspider on 15/9/28.
//  Copyright © 2015年 fff. All rights reserved.
//

import Foundation
import Alamofire
import SwiftyJSON
import OAuthSwift

struct Config {
    static let consumerKey = "" // replace it
    static let consumerSecret = "" // replace it
    static let callbackURL = "oauth-swift://oauth-callback/twitter"
}

[...]

Issue #115, opened 2015-09-29
Error in signature calculation


olivier38070:
I ccompared with that web site : 
http://oauth.googlecode.com/svn/code/javascript/example/signature.html

the signature calculated in OAuthSwiftClient, authorizationHeaderForMethod give wrong value.
this imply that on some web site, upload fail.
here is something I did, which works.

hope this help.
Olivier

    // convert a Dictionary, in array of string, with escaped string
    func getEscapedArrayParam(parameters: Dictionary<String, AnyObject>) -> [String] {
        // escape all oauth parameter
        var encodedParam = [String]()
        for(k, v) in parameters {
            let str = k + "=" + (v as! String)
            let escapedStr = str.stringByAddingPercentEncodingWithAllowedCharacters(.URLHostAllowedCharacterSet())
            encodedParam.append(escapedStr!)
        }
        // sort the result
        encodedParam.sortInPlace{ $0 < $1 }
        return encodedParam
    }
    
    // do almost like stringByAddingPercentEncodingWithAllowedCharacters, but take in account more character to be changed in %xyz
    //!*'();:@&=+$,/?%#[]
    func escapeString(str: String) -> String {
        let encoding: NSStringEncoding = NSUTF8StringEncoding
        // !*'();:@&=+$,/?%#[]
        let charactersToBeEscaped = ":/?&=;+!@#$()',*" as CFStringRef
        let charactersToLeaveUnescaped = "[]." as CFStringRef
        let raw: NSString = str
        let result = CFURLCreateStringByAddingPercentEscapes(kCFAllocatorDefault, raw, charactersToLeaveUnescaped, charactersToBeEscaped, CFStringConvertNSStringEncodingToEncoding(encoding))
        return result as String
    }
    
    
    public func encodeParameters(method: String, url: NSURL, parameters: Dictionary<String, AnyObject>, credential: OAuthSwiftCredential) -> String {
        
        // algorithm used : https://dev.twitter.com/oauth/overview/creating-signatures
        // will add the oauth_signature parameter
        
        //define the oauth parameters
        var authorizationParameters = Dictionary<String, AnyObject>()
        authorizationParameters["oauth_version"] = OAuth.version
        authorizationParameters["oauth_consumer_key"] = credential.consumer_key
        authorizationParameters["oauth_timestamp"] = String(Int64(NSDate().timeIntervalSince1970))
[...]

PR #165, opened 2015-12-26
fix urlEncodedStringWithEncoding to remove reserved characters

tkawaji:
There is signiture error when parameter value includes line feed code(\n).
So I basically use NSCharacterSet.URLQueryAllowedCharacterSet for stringByAddingPercentEncodingWithAllowedCharacters.
And remove some characters defined as reserved in RFC 3986.

I think there are big changes in urlencoding behavior from existing source code.
So I would appreciate if you could check carefully.

Thank you.

### DIFF
--- a/OAuthSwift/String+OAuthSwift.swift
+++ b/OAuthSwift/String+OAuthSwift.swift
@@ -33,7 +33,11 @@ extension String {
 
     func urlEncodedStringWithEncoding(encoding: NSStringEncoding) -> String {
         let originalString: NSString = self
-        let customAllowedSet =  NSCharacterSet(charactersInString:" :/?&=;+!@#$()',*=\"#%/<>?@\\^`{|}").invertedSet
+        // Reserved characters in RFC 3986
+        let genDelims = ":/?#[]@"
+        let subDelims = "!$&'()*+,;="
+        let customAllowedSet = NSCharacterSet.URLQueryAllowedCharacterSet().mutableCopy() as! NSMutableCharacterSet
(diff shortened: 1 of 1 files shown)


phimage:
yes it's a thing to look carefully

there is some code here also #115 

And after studying the case of space character here #156, I read in RFC that there is two things
Encoding for signature and for further url request

Did you try to launch unit test using the workspace and pod command to retrieve dependencies?
there is a signatureTest for only one case

tkawaji:
I couldn't setup unit test environment.
Is there any document about how to launch unit test?

phimage:
No sorry, I must do it one day

Cocoapod must be installed
```
sudo gem install cocoapods
```
(close any xcode OAuthSwift project if opened)
in OAuthSwift directory 
```
 pod install
```
open OAuthSwift.xcworkspace (not OAuthSwift.xcodeproj)

To launch test, choose scheme OAuthSwiftTests and do Menu:Product then Test (Command U)

phimage:
Just tested `testSignature` and it's ok

tkawaji:
Thank you.

I setup test environment, but I have some error:
"HttpResponseBody has no member STRING"

I think that version of Swifter is different.
need Podfile.lock?

phimage:
Swifter was 1.0.2 instead of 1.0.6 on my repository
I fix that, thanks : b8190756e87994553233f9b6f9f9ee6ea607a261

tkawaji:
Thank you. I could run test case.
And testSignature is OK, too.

phimage:
PR #201 merged

PR #168, opened 2016-01-01
URL-Encode square brackets Fix #174

### LINKED ISSUE #174: Incorrect signature when parameter includes a [ or ]
Using the AWeber public API (https://labs.aweber.com/docs) I discovered that when passing parameters with OAuthSwift requests that include braces, [ or ], they aren't actually encoded, and thus the signature isn't properly created.  When the server attempts to verify the signature, it does so based on encoded values for the braces, and thus does not match what the client provides.  I created #173 to fix this.  Let me know what you think.

nighthawk:
I've hit an issue where the OAuth signature was calculated wrongly, which I traced back to some of the parameters of the API that I'm using ([UserVoice](https://developer.uservoice.com/docs/api/reference/)) having parameters such as `object[field]=value` and OAuthSwift didn't escape the square brackets.

I checked this against [the nouncer.com calculator](http://nouncer.com/oauth/signature.html). With my fix, the signature from OAuthSwift matches that, and it's also matching the API that I'm connecting to.

### DIFF
--- a/OAuthSwift/String+OAuthSwift.swift
+++ b/OAuthSwift/String+OAuthSwift.swift
@@ -33,7 +33,7 @@ extension String {
 
     func urlEncodedStringWithEncoding(encoding: NSStringEncoding) -> String {
         let originalString: NSString = self
-        let customAllowedSet =  NSCharacterSet(charactersInString:" :/?&=;+!@#$()',*=\"#%/<>?@\\^`{|}").invertedSet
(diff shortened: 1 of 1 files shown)


nighthawk:
Note, I tried the fork from the currently open pull request #165, but that did not fix my issue.

phimage:
Thanks for the details
I see brackets in genDelims  var from #165, strange

To merge some PR about signature and url encoding, I need to add more tests into OAuthSwiftTests/SignTests.swift with valid dataset
I will take a look at 'the nouncer.com calculator" and http://oauth.googlecode.com/svn/code/javascript/example/signature.html

phimage:
PR #201 merged

PR #177, opened 2016-01-24
Fix directory not found warning

toohotz:
Hey there, I've noticed that inside of the project there's a library search path specified for ```"$(SDKROOT)/usr/lib/system"```.  Seems like in Xcode 7, the search paths flags have changed and removing that specified search path fixes the warning (also checked to make sure nothing else changed in the process).  I already made the change on my local copy of the OAuthSwift project.

This occurs when building for device, does not occur when building for sim.


phimage:
`LIBRARY_SEARCH_PATHS` is not added when creating a new project
I trust you and merge your PR, thanks

PR #186, opened 2016-02-18
made oauth_verifier public on OAuthSwiftCredential

itchingpixels:
Hi, 
Thanks for the awesome work again! We found OAuthSwift super useful.

We have been working on a project where the backend requires us to send the "oauth_verifier" to the server after a user has authenticated.
I consulted with the backend guys and they said this requirement is coming from the official Flickr library they're using.
Would it be alright to expose this property as public? Are there any other workarounds we could use to access it without necessarily polluting the public API? Has this been requested before?

Thanks a lot!

### DIFF
--- a/OAuthSwift/OAuthSwiftCredential.swift
+++ b/OAuthSwift/OAuthSwiftCredential.swift
@@ -49,7 +49,7 @@ public class OAuthSwiftCredential: NSObject, NSCoding {
     var consumer_secret: String = String()
     public var oauth_token: String = String()
     public var oauth_token_secret: String = String()
-    public var oauth_verifier: String = String()
+    public internal(set) var oauth_verifier: String = String()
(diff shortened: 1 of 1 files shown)


phimage:
agree to expose (and thanks to detail your needs)

maybe add `private(set)` before `public`
This allow to make variable readonly

itchingpixels:
sounds like a good idea, but I came across a problem: since private(set) restricts it for the given file, I'll get this error when I try to run the tests:

image

itchingpixels:
internal(set) works, that restricts it to be read-only outside of the framework, is that good?

phimage:
yes put internal set and I merge

itchingpixels:
done :+1:

itchingpixels:
thanks!

PR #188, opened 2016-02-20
Add an option to send the OAuth1.0 parameters as request parameters (fix #187)

### LINKED ISSUE #187: Withings server does not process the oauth parameters sent in Authorization header
If we send a request to Withings API to access any resources, the server returns an error : 

```
{
    body =     {
    };
    error = "Not enough params";
    status = 503;
}
```
I have used the following code to send the request with the demo app :
```
func doOAuthWithings(serviceParameters: [String:String]){
        let oauthswift = OAuth1Swift(
            consumerKey:    serviceParameters["consumerKey"]!,
            consumerSecret: serviceParameters["consumerSecret"]!,
            requestTokenUrl: "https://oauth.withings.com/account/request_token",
            authorizeUrl:    "https://oauth.withings.com/account/authorize",
            accessTokenUrl:  "https://oauth.withings.com/account/access_token"
        )
        oauthswift.authorizeWithCallbackURL( NSURL(string: "oauth-swift://oauth-callback/withings")!, success: {
            credential, response, parameters in
            self.showTokenAlert(serviceParameters["name"], credential: credential)
            self.testWithings(oauthswift, userId: parameters["userid"]!)
            }, failure: { error in
                print(error.localizedDescription)
        })
    }
    func testWithings(oauthswift: OAuth1Swift, userId : NSString) {
        //oauthswift.client.parameterTransmissionMethod = .RequestURIQuery
        oauthswift.client.get("https://wbsapi.withings.net/v2/measure", parameters: ["action":"getactivity", "userid":userId, "date":"2016-02-15"],
            success: {
                data, response in
                let jsonDict: AnyObject! = try? NSJSONSerialization.JSONObjectWithData(data, options: [])
                print(jsonDict)
            }, failure: { error in
                print(error.localizedDescription)
        })
    }
```
It appears that the Withings server does not process the oauth parameters sent in Authorization header but wait for them in the request parameters.
It is not usual but it is compliant with the [RFC 5849](https://tools.ietf.org/html/rfc5849#section-3.5) which defines 3 differents ways to transmit the oauth parameters.

I have forked the repository to make a fix and I will submit a PR soon.

- **OS targeted (with version):** iOS 9.2
- **OAuth provider:** Withings (OAuth1.0)
- **OAuthSwift version:** master head

jdrevet:
As described in issue #187, some servers (at least the Withings one) require that the OAuth1.0 parameters are transmitted as query parameters.

As it is not an usual behaviour, I have tried to minimize the changes by using convenience initializers and optionnals parameters with default value.

There is 2 places where the process differs: 
- In OAuthSwiftClient where the oauth parameters are set in the authorization header or in query paremeters
- In OAuthSwiftHTTPRequest where the oauth parameters should not be removed from the query parameters in the new case

### DIFF
--- a/OAuthSwift/OAuth1Swift.swift
+++ b/OAuthSwift/OAuth1Swift.swift
@@ -22,7 +22,7 @@ public class OAuth1Swift: OAuthSwift {
     var access_token_url: String
     
     // MARK: init
-    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String, paramsLocation : OAuthSwiftHTTPRequest.ParamsLocation){
+    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String){
         self.consumer_key = consumerKey
         self.consumer_secret = consumerSecret
         self.request_token_url = requestTokenUrl
@@ -30,15 +30,6 @@ public class OAuth1Swift: OAuthSwift {
         self.access_token_url = accessTokenUrl
         super.init(consumerKey: consumerKey, consumerSecret: consumerSecret)
         self.client.credential.version = .OAuth1
-        self.client.paramsLocation = paramsLocation;
-    }
-    
-    public convenience init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String) {
-        self.init(consumerKey:consumerKey, consumerSecret: consumerSecret,
-            requestTokenUrl: requestTokenUrl,
-            authorizeUrl: authorizeUrl,
-            accessTokenUrl: accessTokenUrl,
-            paramsLocation: OAuthSwiftHTTPRequest.ParamsLocation.AuthorizationHeader)
     }
 
     public convenience init?(parameters: [String:String]){
--- a/OAuthSwift/OAuth1Swift.swift
+++ b/OAuthSwift/OAuth1Swift.swift
@@ -22,7 +22,7 @@ public class OAuth1Swift: OAuthSwift {
     var access_token_url: String
     
     // MARK: init
-    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String){
+    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String, paramsLocation : OAuthSwiftHTTPRequest.ParamsLocation){
         self.consumer_key = consumerKey
         self.consumer_secret = consumerSecret
         self.request_token_url = requestTokenUrl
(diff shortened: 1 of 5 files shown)


OAuthSwift/OAuth1Swift.swift
```diff
@@ -22,7 +22,7 @@ public class OAuth1Swift: OAuthSwift {
     var access_token_url: String
     
     // MARK: init
-    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String){
+    public init(consumerKey: String, consumerSecret: String, requestTokenUrl: String, authorizeUrl: String, accessTokenUrl: String, paramsLocation : OAuthSwiftHTTPRequest.ParamsLocation){
```
phimage: instead of adding new `init`, default value could be used
`, paramsLocation : OAuthSwiftHTTPRequest.ParamsLocation = .AuthorizationHeader`

OAuthSwift/OAuthSwiftClient.swift
```diff
@@ -10,9 +10,10 @@ import Foundation
 
 var OAuthSwiftDataEncoding: NSStringEncoding = NSUTF8StringEncoding
 
-public class OAuthSwiftClient {
+public class OAuthSwiftClient: NSObject {
```
phimage: why OAuthSwiftClient must extends NSObject ?

OAuthSwift/OAuth1Swift.swift
jdrevet: I agree that it would be right to let users to set the paramsLocation directly. In this way, we do not modify OAuth1Swift for an unusal behaviour.

OAuthSwift/OAuthSwiftClient.swift
jdrevet: I have added NSObject subclassing to allow OAuthSwiftClient to be called from Objective-C. Most other classes inherit from NSObject (in particular OAuthSwift, OAuthSwiftCredential and OAuthSwiftHTTPRequest), so I have thought that it would not be a non sense to do that.

phimage: ok I take it
next time another PR and commiy

jdrevet: Ok, thank you. 
It is my first significant contribution to a projet on Github so I do not have the right habits yet.

OAuthSwift/OAuthSwiftHTTPRequest.swift
```diff
@@ -54,7 +60,7 @@ public class OAuthSwiftHTTPRequest: NSObject, NSURLSessionDelegate {
         self.init(URL: URL, method: .GET, parameters: [:])
     }
 
-    init(URL: NSURL, method: Method, parameters: Dictionary<String, AnyObject>) {
+    init(URL: NSURL, method: Method, parameters: Dictionary<String, AnyObject>, paramsLocation : ParamsLocation? = .AuthorizationHeader) {
```
phimage: `ParamsLocation?` could become `ParamsLocation` and so no need to do unwrap : `paramsLocation!`
unwrap is dangerous

in `setupRequestForOAuth` and `makeRequest` too

phimage:
thanks for the contribution
I think I add my last comment, I review all the code.
Fix and I will merge

PR #192, opened 2016-02-24
Use custom authorisation scheme

Poupouch:
We experienced some bug using LinkedIn OAuth and the "Authorization" header. Turns out this bug disappear if we use the parameter "oauth2_access_token" instead. This brings us to think about this correction.

I want to be able to use my own authorisation method using custom params. See LinkedIn Demo for example.


OAuthSwiftDemo/ViewController.swift
```diff
@@ -408,14 +408,15 @@ extension ViewController {
         oauthswift.authorizeWithCallbackURL( NSURL(string: "http://oauthswift.herokuapp.com/callback/linkedin2")!, scope: "r_fullprofile", state: state, success: {
             credential, response, parameters in
             self.showTokenAlert(serviceParameters["name"], credential: credential)
-            self.testLinkedin2(oauthswift)
+            self.testLinkedin2(credential.oauth_token, oauthswift: oauthswift)
```
phimage: no need to pass token to test method, could be retrieved from oauthswift object
best to keep same signature for all test method if possible (for easy copy paste)

OAuthSwift/OAuthSwiftClient.swift
```diff
@@ -122,13 +122,14 @@ public class OAuthSwiftClient: NSObject {
             case .RequestURIQuery:
                 //Add oauth parameters as request parameters
                 request.parameters += self.credential.authorizationParametersWithSignatureForMethod(method, url: signatureUrl, parameters: signatureParameters, body: body)
+            case .CustomParameters:
+                break
```
phimage: If I understand this is not Custom but no oauth parameters
A Custom will be something like `.CustomParameters([String:String])` or maybe dynamic one with closure `.CustomParameters(@autoclosure () -> [String:String])` (never tested)

phimage:
I can accept to merge a PR with 6 commits with only one (see contributing guide)

Use git to make only one commit, use rebase instead of pull when possible, etc...
_On way is to create a branch from previous commit named feature-xxxx, rebase our master branch into this new branch, then apply a patch to this branch and commit...._

Then I am not sure this is the solution for your problem
Adding "oauth2_access_token" for each request is weird
Maybe `.CustomParameters(["oauth2_access_token": token])` will be better

(I think of other solution, allow to customise the string used internally if possible)

Poupouch:
Just saw the commit mess, willl create an other PR, thanks.

In our case we use the function only once so that make sens to me but I see the struggle, will think about something else. Thank you!

PR #194, opened 2016-02-24
OAUTH_APP_EXTENSIONS  inside UIApplication extension (Fixes #193) + remove & from query string 

### LINKED ISSUE #193: SharedApplication is unavailable (UIApplication+OAuthSwift.swift)
I get the following compiler-error:
```
UIApplication+OAuthSwift.swift:14:34: 'sharedApplication()' is unavailable: Use view controller based solutions where appropriate instead.
```

It refers to the following code from OAuthSwift:
```
static var topViewController: UIViewController? {
    return UIApplication.sharedApplication().topViewController
}
```

Of course I set `OAUTH_APP_EXTENSIONS`..

Everywhere else in the code, OAuthSwift uses:
```
#if !OAUTH_APP_EXTENSIONS
    UIApplication.sharedApplication()......
#endif
```
Is there a reason this if-statement is not used inside `UIApplication+OAuthSwift.swift`?

(Newest OAuthSwift version, installed with CocoaPods, but I also tried installing manually)

- **OS targeted (with version):** iOS 9.0
- **OAuth provider:** -
- **OAuthSwift version:**  0.5.1

Mauricevb:
* Make sure OAUTH_APP_EXTENSIONS works inside the UIApplication extension too (Fixes #193)
* Remove &-symbol from keys in query string. Query strings can start with an & so. Before this fix some keys started with an &.

### DIFF
--- a/OAuthSwift/String+OAuthSwift.swift
+++ b/OAuthSwift/String+OAuthSwift.swift
@@ -58,7 +58,8 @@ extension String {
             key = nil
             scanner.scanUpToString(keyValueSeparator, intoString: &key)
             scanner.scanString(keyValueSeparator, intoString: nil)
-
+			key = key?.stringByReplacingOccurrencesOfString("&", withString: "")
+			
             value = nil
             scanner.scanUpToString(elementSeparator, intoString: &value)
             scanner.scanString(elementSeparator, intoString: nil)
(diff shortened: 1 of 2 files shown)


phimage:
I will not merge immediately because your PR contains two features
( one feature by PR,  see guide https://github.com/OAuthSwift/OAuthSwift/blob/master/CONTRIBUTING.md , you can make branch for each )

So I have question on your second fixe, any example? string and value which failed (need to know also the oauth provider which failed)
(With this I can create unit test maybe ;p )

PR #196, opened 2016-02-24
Remove first char from query string when needed

Mauricevb:
I use OAuthSwift with an OAuth 1 API. The API returns a query string which starts with an &-symbol. Because of this, the first key will be parsed as '&oauth_token'.
This pull request fixes this problem.


Mauricevb:
This pull request should only contain commit 79eded3... I don't know what commit 9c917d9 does here. I give up, I'll learn forking before contributing again .

phimage:
Ok, pull request take all commit from a branch

PR #197, opened 2016-02-24
Remove first char from query string when needed

Mauricevb:
I use OAuthSwift with an OAuth 1 API. The API returns a query string which starts with an &-symbol. Because of this, the first key will be parsed as '&oauth_token'.
This pull request fixes this problem.


phimage:
thanks

PR #198, opened 2016-02-25
Custom URI parameter for oauth token

Poupouch:
You can chose a custom parameter name for oauth_token.

Inspired by a LinkedIn OAuth common error using the Authorization header.

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -122,13 +122,15 @@ public class OAuthSwiftClient: NSObject {
             case .RequestURIQuery:
                 //Add oauth parameters as request parameters
                 request.parameters += self.credential.authorizationParametersWithSignatureForMethod(method, url: signatureUrl, parameters: signatureParameters, body: body)
+            case .OAuthTokenParameter(let param):
+                //Add oauth token as custom parameter
+                request.parameters[param] = self.credential.oauth_token
             }
             
             if let addHeaders = headers {
                 requestHeaders += addHeaders
             }
             request.headers = requestHeaders
-
             request.dataEncoding = OAuthSwiftDataEncoding
             return request
         }
--- a/OAuthSwift/OAuthSwiftHTTPRequest.swift
+++ b/OAuthSwift/OAuthSwiftHTTPRequest.swift
@@ -21,8 +21,8 @@ public class OAuthSwiftHTTPRequest: NSObject, NSURLSessionDelegate {
         }
     }
     
-    @objc public enum ParamsLocation : Int {
-        case AuthorizationHeader, /*FormEncodedBody,*/ RequestURIQuery
+    public enum ParamsLocation {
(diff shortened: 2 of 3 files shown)


phimage:
Thanks to investigate a more clean solution for your specific problem

Now that you use a specific enum with parameter as I ask to have a more clean code, no more `@objc ` on enum `ParamLocation`
maybe @jdrevet do it at purpose ;(

I must think about it before merging

Poupouch:
According to the commit ([ce1f82bf](https://github.com/OAuthSwift/OAuthSwift/commit/ce1f82bf8d6aee6003720ad901831dd4090a3cbc)) introducing `ParamLocation`, I don't see any reason to keep the `@objc` but I may miss something. Also the use of the `ParamLocation` parameter is now reduced to its class of definition `OAuthSwiftHTTPRequest` and where it is instantiated `OAuthSwiftClient`.

Thanks for the reviewing. :)

jdrevet:
I have added the @obj and Int typing for ParamLocation enum to be able to manipulate it from Objective-C project. But I understand that Objective-C compatibility should not slow down the enhancements of the library.
Nevertheless, I am not sure if the ParamLocation enum is the right place to configure a custom parameter name. Originally, It has been imagined to represent the 3 ways defined in the OAuth1.0 RFC  https://tools.ietf.org/html/rfc5849#section-3.5 to transmit the OAuth parameters.

Poupouch:
According to the [OAuth2.0](https://tools.ietf.org/html/rfc6749#page-50) documentation, you can use those "custom" parameters. Since the `OAuthSwiftHTTPRequest` is the same for OAuth1 and OAuth02, I couldn't think about another quick solution.

phimage:
ok I think I will work on other solution
add `public var globalParams: [String: AnyObject]? = nil` to `OAuthSwiftClient` 

and somewhere this params will be added to all the requests...
```
            if let global = globalParams {
                request.parameters += global
            }
```
...before or after the `switch` on `paramsLocation`...

Then there is some issue on parameters to filters maybe, because `RequestURIQuery`will not do all the trick (url will be filled also with standards parameters and cause issues for LinkedIn)
...
I keep open this PR until a solution is found

Poupouch:
The spec for `OAuth1` and `OAuth2` are different concerning the `paramsLocation` specification. For `OAuth1` you'll have Authorization Header, Form-Encoded Body and Request URI Query and for `OAuth2` it will depend on the type of the Access token (considering [this](https://tools.ietf.org/html/rfc6749#section-7)).

Can we consider separate the `paramLocation` for the 2 version, so move the actual `paramLocation` to `OAuth1Swift` and having something more general for `OAuth2Swift` (I would say Authorization Header, Form-Encoded Body and Custom URI Query) ?

Poupouch:
This issue has been perfectly resolved by using a headerFactory and custom parameters. I think we can close it.

phimage:
ok thanks

PR #200, opened 2016-03-02
Add Trello and Buffer

alexandreraulin:
Hi, I use OAuthSwift for Trello and Buffer so I have added examples to the demo app.

Alexandre


phimage:
thanks for the contribution

PR #202, opened 2016-03-09
Added a Goodreads example to the demo app.

danielbmarques:


phimage:
thanks for this PR but there is 3 commits for one things (git is weird some times)
but to keep history clean could you make another PR with only one commit
you can try to "merge" the commit (with git tricks...)
or faster I think, create a branch feature-goodreads before your commits, apply a patch and commit, make a PR with the branch

danielbmarques:
Ok, there you go. #204

PR #203, opened 2016-03-10
Feature goodreads

danielbmarques:
Added a Goodreads example to the demo app.


danielbmarques:
Argh, still two commits! Sorry...

PR #210, opened 2016-04-01
Fixed Swift 2.2 build warnings

lyptt:
This change fixes build warnings in the OAuthSwift target when using the latest version of Xcode.


phimage:
I think converting a for to a while must move the increment/modification code at the end of the block
Or the first case will be missing

lyptt:
Good spot @phimage on the while issue, this should be fixed now.

Issue #211, opened 2016-04-01
Provide a cancel function on OAuthSwiftHTTPRequest


FGoessler:
It would be nice to cancel running HTTP requests. NSURLSessionTask offers a simple API for that (just call cancel), but since the NSURLSessionTask is not exposed in the OAuthSwiftHTTPRequest class we cannot access it.

One could either provide a method on OAuthSwiftHTTPRequest to cancel the request or just expose the underlying task as a property.

phimage:
Add near `session` an attribute `task` into `OAuthSwiftHTTPRequest` and provide the `stop`or `cancel` function (function which do nothing if no task)

or alternatively add a callback on `start` method to get the `task`...

You can PR

PR #213, opened 2016-04-03
Added cancel method on OAuthSwiftHTTPRequest

FGoessler:
With this pull request I implemented a cancel a method on OAuthSwiftHTTPRequest to allow clients to cancel request easily, as discussed in #211.

I also added a unit test for this new functionality but discovered that some other tests are failing on my machine - before my change as well as after my change and also for the last released version 0.5.1. I haven't fixed them because I don't know why they are broken and this is not related to this feature. Maybe some issue with Xcode 7.3? Haven't tried with older Xcode versions.


phimage:
Yes there is experimental unit tests and failure..

Thx for the pr

PR #215, opened 2016-04-03
Provide a method on OAuthSwiftClient that accepts a plain NSURLRequest Fix #212.

### LINKED ISSUE #212: Provide a method on OAuthSwiftClient that accepts a plain NSURLRequest
To interact with existing Objective C APIs, which often work with NSURLRequests, it would be nice to have a method on OAuthSwiftClient which accepts an NSURLRequest in addition to the more convenient method that accept URL string, parameters, headers and co.

With that it would also be possible to use advanced properties of NSURLRequest like cachePolicy and timeoutInterval.

FGoessler:
Added a method on OAuthSwiftClient that accepts a plain NSURLRequest, creates an OAuthSwiftHTTPRequest object with it and starts that request as usually, like discussed in #212.

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -72,78 +72,86 @@ public class OAuthSwiftClient: NSObject {
         }
     }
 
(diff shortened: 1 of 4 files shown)


phimage:
well done, thx

Issue #214, opened 2016-04-03
Fix unit tests to improve trust in this library


FGoessler:
We want to use this library in a quite big project in our efforts to migrate our network stack to Swift and OAuth 2 from a custom build OAuth 1 lib written in Objective C. Since OAuthSwift would then be a big essential part of our app we need a very high reliability in it's functionality and quality.

Unfortunately some unit tests of this project are currently failing and there also some that are just commented out and ignored.

To increase the trust in this library it would be nice to have 100% running unit tests with a good coverage which are also automatically run against each pull request e.g. via Travis.

phimage:
Yes more unit test and travis are in roadmap https://github.com/OAuthSwift/OAuthSwift/wiki/Roadmap
I ask authorization for travis (organisation owner must accept it)

I have not run unit test since previous release.
There is many change into signature for OAuth1.

FGoessler:
Imo the following tasks should be done:

- [ ] Fix existing unit tests
- [ ] Look at all test that are commented out and fix them or delete them
- [ ] Setup Travis CI for this repo and let it check PRs
- [ ] Test previously untested stuff:
   - [ ] Multipart requests
   - [ ] NSDate category
   - [ ] String category
   - [ ] Utils, SHA1 and HMAC
   - [ ] Access Token renewal
   - [ ] Convenience & request methods on OAuthSwiftClient

phimage:
some unit tests has been added

PR #218, opened 2016-04-05
Fix and improve unit tests

FGoessler:
This PR aims to fix and improve the unit tests as discussed in #214, although it does not address all issues mentioned there.


phimage:
good work, thanks

Issue #217, opened 2016-04-04
Refresh OAuth2 access tokens by default in the OAuthSwiftClient


FGoessler:
@fabiomassimo created a convenience method on `OAuth2Swift` to automatically refresh the access token in case of a failure for a normal request in #209 which is nice 👍.

I expected this to be the default behaviour of every request made via the `OAuthSwiftClient`, which seems not to be the case right now.

My suggestion:
- Put the logic for the refresh in the `OAuthSwiftClient` and only execute it for OAuth2 requests. 
- Which OAuth version is used can be easily extracted from the `OAuthSwiftCredential` of the client.
- Since the convenience method currently requires some other methods which are implemented on `OAuth2Swift` I would either move them as well, move them to an extension of `OAuthSwiftClient` in the `OAuth2Swift` file or transform them to class methods. What is the preferred solution?
- Mark the convenience on `OAuth2Swift` as deprecated and let it just call through to the client. Or even remove the convenience method since no release was made since the introduction of the method and therefore no consumer of this library should be affected.

Any thoughts, comments, considerations, suggestions, ...?

I would start working on a PR then. 😉

antwerpenR:
I am trying to get this to work with Meetup.com and it seems that the access token is correctly set in line 83 of authorizeWithCallbackURL but you are not storing the "expires_in" responseParameter.  Also, I can find no trace of any possible use of "grant type=refresh-token" and requestOAuthAccessTokenWithParameters on line 160 is never being called.   For reference, the Meetup Spec is here:  http://www.meetup.com/meetup_api/auth/#oauth2server-auth-success

Am I doing something wrong or is this a bug?  It does seem that the token is correct and signing of requests works properly....just no handling of the refresh

antwerpenR:
I think the solution may just be to store the "expires_in" at the same time that you store the token.   Inserting these lines: 85, 86 and 87.
            if let expiresIn:String = responseParameters["expires_in"], offset = Double(expiresIn)  {
              self.client.credential.oauth_token_expires_at = NSDate(timeInterval: offset, sinceDate: NSDate())
            }

FGoessler:
I assume you're using the current version on master, right?

The expires in value is actually stored in the ```requestOAuthAccessTokenWithParameters``` method.

[...]

PR #219, opened 2016-04-05
Autorenew OAuth2 access tokens

FGoessler:
This PR aims to make the access token renewal process transparent to clients of this library as discussed in #217. 

The implementation differs from my proposed solution in the issue since the token renewal process is to tightly coupled with the `OAuth2Swift` class. 
Now the `OAuthSwiftClient` exposes a callback property which can be set to handle expired/invalid tokens and will retry the request via the completion block. `OAuth2Swift` configures this property such that it requests a new access tokens with its refresh token and stores the new token.

In case a request failed the error response is checked if the cause is an invalid/expired token. Therefore I created an extension on NSError which I based on the code and specification in the [Wiki](https://github.com/OAuthSwift/OAuthSwift/wiki/OAuth-2.0-Token-Expiration). It also handles special cases for Facebook.

I would appreciate feedback about the convenience method `startAuthorizedRequest(...)` in `OAuth2Swift` which was introduced by @fabiomassimo. Since there was no release since this method was added it should be save to remove it, but for now I left it in and marked  it as deprecated.

This PR also includes the changes to the unit tests done in #218 since I rebased this branch onto the other. I can perform a rebase on master as soon as the other PR is merged to make the merge and history nicer.

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -12,12 +12,22 @@ var OAuthSwiftDataEncoding: NSStringEncoding = NSUTF8StringEncoding
 
 public class OAuthSwiftClient: NSObject {
 
-    private(set) public var credential: OAuthSwiftCredential
-    public var paramsLocation: OAuthSwiftHTTPRequest.ParamsLocation = .AuthorizationHeader
-    public var tokenRenewedHandler: OAuthSwift.TokenRenewedHandler?
-    public var tokenExpirationHandler: OAuthSwift.TokenExpirationHandler = { _, completion in
+    public static let noopTokenExpirationHandler: OAuthSwift.TokenExpirationHandler = { _, completion in
         completion(error: NSError(domain: OAuthSwiftErrorDomain, code: OAuthSwiftErrorCode.TokenExpiredError.rawValue, userInfo: nil))
     }
+
+    private(set) public var credential: OAuthSwiftCredential
+    public var paramsLocation: OAuthSwiftHTTPRequest.ParamsLocation = .AuthorizationHeader
+
+    /// This handler gets called when the access token is successfully renewed via the tokenExpirationHandler.
+    public var tokenRenewedHandler: OAuthSwift.TokenRenewedHandler?
+
+    /// This handler gets called when the OAuth2 access token has expired and gives a chance to 
+    /// refresh it. The request will be tried again if the completion is called without an error.
+    /// Using OAuth2Swift will configure its client with a tokenExpirationHandler by default.
+    ///
+    /// If you don't want a working tokenExpirationHandler use the OAuthSwiftClient.noopTokenExpirationHandler.
+    public var tokenExpirationHandler: OAuthSwift.TokenExpirationHandler = noopTokenExpirationHandler
 
     static let separator: String = "\r\n"
     static var separatorData: NSData = {
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -12,10 +12,6 @@ var OAuthSwiftDataEncoding: NSStringEncoding = NSUTF8StringEncoding
 
 public class OAuthSwiftClient: NSObject {
 
-    private static let noopTokenExpirationHandler: OAuthSwift.TokenExpirationHandler = { _, completion in
-        completion(error: NSError(domain: OAuthSwiftErrorDomain, code: OAuthSwiftErrorCode.TokenExpiredError.rawValue, userInfo: nil))
-    }
-
     private(set) public var credential: OAuthSwiftCredential
     public var paramsLocation: OAuthSwiftHTTPRequest.ParamsLocation = .AuthorizationHeader
 
@@ -33,11 +29,6 @@ public class OAuthSwiftClient: NSObject {
             )
         }
     }
-
-    static let separator: String = "\r\n"
-    static var separatorData: NSData = {
-        return OAuthSwiftClient.separator.dataUsingEncoding(OAuthSwiftDataEncoding)!
-    }()
 
     // MARK: init
     public init(consumerKey: String, consumerSecret: String) {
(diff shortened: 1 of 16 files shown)


fabiomassimo:
The expiration token details are only **recommended** for OAuth 2.0 ([RFC 6749](https://tools.ietf.org/html/rfc6749#appendix-A.14)).
Extend this scenario to the `OAuthClient` might hide this support by extending it to OAuth 1.0 clients as well. 
What do you think to add an assertion to help the user of the library to not erroneously set the callback properly in case of not supported OAuth version?

Thank you for such great proposal. I liked the simple and efficient way you managed the error with a proper extension over NSError.

FGoessler:
Thanks - that's a good point - I'll add such an assertion in the client, when trying to set a ```tokenExpirationHandler```.

phimage:
Thanks for the great contribution
I have no time to merge now but I will check later

I think the wiki page https://github.com/OAuthSwift/OAuthSwift/wiki/OAuth-2.0-Token-Expiration is now obsolete
Maybe README.md must be edited or wiki page must be modified or removed

FGoessler:
Yes - updating those markdown documents is something I should definitely do. 👍

Beside this I discovered some other issues, on which I'm currently working but which are only partially related to this PR since some were already existing before this PR:
- Calling cancel on an ```OAuthSwiftHTTPRequest``` has no effect if the start method was not called yet or if the cancel method is called in the same run loop/ immediately after calling start. This is due to the async behaviour of the start method.
- Canceling an request that automatically renews the access token is not possible right now, since such an request can consist of up to 3 requests. We should provide a handle there to cancel the complete process no matter in which phase it is, too. 
- After inspecting the code further I think that sending an ```OAuthSwiftHTTPRequest```, which was created with an NSURLRequest, will not work properly since headers, params and co are not updated in the request correctly. The problem here is that a ```OAuthSwiftHTTPRequest``` tracks it's configuration (URL, params, headers, ...) possibly twice - in local properties as well as in a ```NSURLRequest``` object. You can modify one without the other being updated - which is done when making a request right now. I'm currently thinking about several solutions here, but most of them involve a lot of changes and refactoring - I'll create an PR or issue later when I have time and a solution in sight. I'm mostly thinking about having some sort of immutable config object for requests, which can be passed around, and centralising the request building somewhere. Right now parts of the request are configured in the ```OAuthSwiftHTTPRequest``` and other parts are configured in the ```OAuthSwiftClient```. 🤔

FGoessler:
I did some refactoring and fixed the three issues that I mentioned above.

From my commit message:
> - Created a new immutable OAuthSwiftHTTPRequestConfig struct to encapsulate the configuration of a request.
- Moved the logic to create multipart requests into that new struct.
- OAuthSwiftHTTPRequest now encapsulates the state of one specific request and is responsible for converting an OAuthSwiftHTTPRequestConfig to a request which has all oAuth specific parameters/headers and which can actually be executed.
- Moved all complex request creation logic from the OAuthSwiftClient to either the OAuthSwiftHTTPRequestConfig or the OAuthSwiftHTTPRequest.
- Created the OAuthSwiftTokenRefreshingRequest which encapsulates the access token refreshing process.

The ```OAuthSwiftClient``` now returns a ```OAuthSwiftRequestHandle``` to cancel the request.

Beside this I think I found a bug when building multipart request. There was an "r\n" sequence but I think it should be an "\r\n". See [a1b0a38](https://github.com/OAuthSwift/OAuthSwift/pull/219/commits/a1b0a384607a85651d6e2286c412c2fe63a62843). I created some unit test for the multipart request creation as well, since that part wasn't tested at all.

I think there might still be some bugs around multipart requests, but I didn't had time to look for an actual specification or debug all possible cases. 

One issue I have is with the ```parameters``` parameter that gets passed in to the client and co.
- For requests without body (GET, HEAD, ...) those parameters are put in the query string.
- For requests that support a body (POST, PUT, PATCH) the parameters are encoded in the body either as string or as JSON based on existing content-type headers.
- I don't know what happens or should happen when I create a request with explicit body and parameters?
- I don't know what happens or should happen when I create a multipart request with parameters.

Is there any docu or specification about this behaviour?

Currently I'm thinking about creating some "high-level" unit tests for "all possible" request combinations, which I can run against the implementation before this refactoring and after it, but I'm not sure if the behaviour was correct before...

OAuthSwift/OAuthSwiftClient.swift
```diff
@@ -10,15 +10,30 @@ import Foundation
...
 
+public protocol OAuthSwiftRequestHandle {
+    var requestConfig: OAuthSwiftHTTPRequestConfig {get}
+    func cancel()
+}
+
 public class OAuthSwiftClient: NSObject {
 
-    private(set) public var credential: OAuthSwiftCredential
+    public var credential: OAuthSwiftCredential
```
phimage: why set public credential? 
(all field are not serialized, this could add a bug if someone set the credential, the protocol version will be 1)

OAuthSwiftTests/OAuthSwiftClientTests.swift
```diff
@@ -23,77 +23,79 @@ class OAuthSwiftClientTests: XCTestCase {
...
-        testMakeNSURLRequest(.GET, url + "?a=a&b=b")
-    }
-
-    func testMakePOSTRequest() {
-        testMakeRequest(.POST, url:url, emptyParameters, url, nil)
-        testMakeRequest(.POST, url:url, ["a":"a"], url, ["a":"a"])
-        testMakeRequest(.POST, url:url, ["a":"a", "b":"b"], url, ["a":"a", "b":"b"])
-        testMakeRequest(.POST, url:"\(url)?c=c", ["a":"a", "b":"b"], "\(url)?c=c", ["a":"a", "b":"b"])
-    }
+//    TODO: refactor / fix these tests
```
phimage: this tests doesn't work anymore?
or just the method signature change? and test must be rewritten? (we must keep compatibility)

OAuthSwift/OAuthSwiftHTTPRequest.swift
```diff
@@ -159,98 +123,150 @@ public class OAuthSwiftHTTPRequest: NSObject, NSURLSessionDelegate {
...
+        additionalParameters: Dictionary<String, AnyObject>,
         dataEncoding: NSStringEncoding,
-        body: NSData? = nil,
-        paramsLocation : ParamsLocation = .AuthorizationHeader) throws -> NSMutableURLRequest {
-            for (key, value) in headers {
-                request.setValue(value, forHTTPHeaderField: key)
+        paramsLocation : ParamsLocation = .AuthorizationHeader,
+        credentials: OAuthSwiftCredential) throws -> NSMutableURLRequest {
+
+            // TODO: I think there might be a bug in the handling of the additionalParameters!
```
phimage: Difficult to merge a big PR with many topic (see https://github.com/OAuthSwift/OAuthSwift/blob/master/CONTRIBUTING.md )
There is no issue referenced with some information about this bug

FGoessler:
I know that this PR is currently why too big for a decent merge and has some quality flaws.

We discovered too many issue while trying to port our network stack from AFNetworking to OAuthSwift so that we decided to make all necessary changes on this branch in my fork of the repo as we encounter them. We probably also need to make some additional changes to have access to the progress APIs that NSURLSession provide.

As my time is quite limited and we need to finish our migration since it's blocking product development, I couldn't improve the code here in detail.

One core issue is that OAuthSwift, as it currently is, is more tailored for an easy of use in quite simple cases but seems to be not powerful enough for a complex apps, which need OAuth based requests all the time and require a lot of the NSURLSession feature set.

Therefore I don't know yet if we will stick with OAuthSwift for our app or write our own OAuth2 client implementation tailored to our needs. I also don't know if I'll be able to further improve this PR and contribute back since I cannot really justify doing this during my normal working time anymore - especially when a custom implementation is faster than a complex refactoring here.

Sorry for the inconvenience :/

phimage:
https://github.com/OAuthSwift/OAuthSwift/releases/tag/0.6.0

I try to get all good codes from your PR and commit it without the "Autorenew OAuth2 access tokens"
I try to not change method signatures and make less change, and maybe on next release I will be able to take again some codes...

serejahh:
Any plans to merge the PR?

phimage:
@serejahh there is too much conflict to merge this PR but the functionality ie. "Autorenew OAuth2 access tokens" and relaunch the request will be implemented one day (when I don't know)

You can use `startAuthorizedRequest` on `OAuth2Swift` object

FGoessler:
Closing as no progress is made here.

PR #224, opened 2016-04-19
Two double quotes to two simple quotes

emericspiroux:
Xcode prevent error with " and all is right when i put ' !

Thanks for your work guys


phimage:
There is different double quote
Maybe you use one from TextEdit

emericspiroux:
Maybe, i've just copy/paste the Readme into my podfile on Xcode, but not for the commit. =S

PR #231, opened 2016-05-20
Ability to specify body data for PUT request

adamford:
* Added an optional "body" parameter to the put method and request method
* Added setting the HTTP Body property to the value passed in to the createRequest method
* Added unit test for making sure the body is set

We have a use case where we need to make a PUT request with data in the body of the request. We couldn't find any publicly accessible way to do this currently, so recommending adding the ability to pass body data in to the request methods.

### DIFF
--- a/OAuthSwift/OAuthSwiftClient.swift
+++ b/OAuthSwift/OAuthSwiftClient.swift
@@ -40,8 +40,8 @@ public class OAuthSwiftClient: NSObject {
         self.request(urlString, method: .POST, parameters: parameters, headers: headers, success: success, failure: failure)
     }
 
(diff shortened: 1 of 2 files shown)


phimage:
thx for the PR
Maybe oauth version 1 will not work because of signature (signature I think take into account the body) but you are right, a body parameter must be added...
I will try to merge it after seeing what to do with other PR

adamford:
@phimage sounds good.
We use oauth version 1 and I was able successfully authenticate and send a PUT request with body using this addition.

PhilizSwiss:
Could we do the same for POST request ?

phimage:
@PhilizSwiss 839604660a9309286ffa234614cf205cec830a85
next time don't hesitate to submit a PR

PR #237, opened 2016-06-17
Ability to provide headers for a multi-part request

dparne:
### Issue faced
- The services I am working against are expecting certain headers in the multipart request other than "Content-Type"
- Since ```func postMultiPartRequest``` is public it would be convenient to be able to pass headers to the underlying request function.

### Changes made
- Added optional headers parameter and updated all functions affected by it.

### DIFF
--- a/OAuthSwift/OAuth2Swift.swift
+++ b/OAuthSwift/OAuth2Swift.swift
@@ -70,7 +70,7 @@ public class OAuth2Swift: OAuthSwift {
     }
 
     // MARK: functions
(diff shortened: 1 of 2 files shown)


phimage:
thx for the contribution
because of method renaming (`renewAccesstokenWithRefreshToken`) in other PR, I can't merge the PR

Feel free to create a new PR or merge the change (with the less commit if you could)
If you have no time I will integrate your change with my own commit at the end of the month

dparne:
@phimage Updated the PR. I have noticed that the latest revision has a lot of unit tests failing though. It does not look like they are caused by my changes. Please let me know if they are.

phimage:
thx
I will do a pass on unit test before releasing a new version

PR #238, opened 2016-06-18
Support for Device Flow Authentication

fabiomassimo:
# Summary

Add support for new grant type: [device 1.0](https://tools.ietf.org/html/draft-ietf-oauth-v2-01#section-3.5.3)

# Why

Due to devices with limited capabilities this grant type makes it possible to implement OAuth 2.0 authentication flow easier.


phimage:
thx!

phimage:
Fix some issue with error code if not json

error code was always OAuthSwiftErrorCode.GeneralError.rawValue and not response.statusCode -> unit test show me that

PR #243, opened 2016-06-24
Make method to renew access token public (and fix typo)

nighthawk:
In our application, we store the refresh token locally and get a new access token at a later time, for this we need the `renewAccessTokenWithRefreshToken` to be public. I thought it might be useful for others, too.


phimage:
thx

PR #257, opened 2016-08-04
custom modalPresentationStyle in SafariURLHandler

teaualune:
While using `SafariURLHandler` to implement OAuth flow on an landscaped iPad app, the presented `SFSafariViewController` transients to a new screen from right to left (*i.e.* the default behavior of `UIModalPresentationStyle.FullScreen`). In order to make the view controller behaves like a modal view, it would be great if we can customize behaviors of the view controller created from `factory(url)`.

This PR proposes a naïve way to do this by exposing `modalPresentationStyle` in `SafariURLHandler`, thus we can customize the presented modal style by doing so:

```swift
let handler = SafariURLHandler(viewController: self)
handler.modalPresentationStyle = .FormSheet
oauthswift.authorize_url_handler = handler
```

Or maybe it is better to make a config block/delegate for clients before the controller is presented in `handle(url)`? Thanks in advance!

### DIFF
--- a/OAuthSwift/OAuthSwiftURLHandlerType.swift
+++ b/OAuthSwift/OAuthSwiftURLHandlerType.swift
@@ -75,6 +75,8 @@ import SafariServices
         public var dismissCompletion: (() -> Void)?
         public var delay: UInt32? = 1
         
+        public var modalPresentationStyle = UIModalPresentationStyle.FullScreen
+
         // init
         public init(viewController: UIViewController, oauthSwift: OAuthSwift) {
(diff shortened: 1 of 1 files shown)


phimage:
There is already a config block named 'factory'

teaualune:
@phimage i see, so i can simply subclass `SafariURLHandler` and override `factory(url)` to achieve this. Thanks!

phimage:
No need to override SafariURLHandler
Juste do 
handler.factory = { url in
....
}

PR #264, opened 2016-08-10
Adding in full swift 2.3 support

cpinski:
Changes:
* Ran the convert to swift 2.3 tool in Xcode 8 Beta 5
* Updated to the Xcode recommended settings
* Added a post install script to the podfile to set the SWIFT_VERSION for each pod. Without this the tests won't run.


phimage:
thx

PR #278, opened 2016-09-24
Optional charset

jstart:
Content-Type    application/x-www-form-urlencoded; charset=Optional(utf-8)


phimage:
Thanks

I will not merge just because I am working into master branch merging the swift3.0 branch and I choose to make an extension for that
```swift
extension String.Encoding {
    
    var charset: String {
        let charset = CFStringConvertEncodingToIANACharSetName(CFStringConvertNSStringEncodingToEncoding(self.rawValue))
        return charset as! String
    }
}
```
but if I cannot update master branch until 3 days I will

PR #279, opened 2016-09-26
Swift 3.0 compatibility for watchOS target

adamkornafeld:


phimage:
did you check master branch?

adamkornafeld:
You're right. These are in master already. I absentmindedly believed swift 3 stuff would be in branch named swift3.0 :)

PR #286, opened 2016-10-07
Fix multipart request with some parameters (swift 2.2)

ktakayama:
A request body would be overridden by query parameters with multipart request.

For example:
- SUCCESS: client.postImage(url, parameters: [:], image: imageData, ....
- FAILURE: client.postImage(url, parameters: parameters, image: imageData, ....

just a fix of this issue is in 21e870acca7e169c871e1d15b0df4fd13a7f2086

I will create pull request to master too later :D


phimage:
Thanks for the two PR and to have added unit tests

PR #309, opened 2016-11-10
Remove 'access_token' parameter when renewing OAuth2 token

boyvanamstel:
Some services (tested with Spotify) will fail the request for a new token when the access token is specified both as a parameter and a header. 

This commit removes the access_token field from the parameters when automatically renewing a token.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -261,9 +261,13 @@ open class OAuth2Swift: OAuthSwift {
                     if let renewalCallBack = onTokenRenewal {
                         renewalCallBack(credential)
                     }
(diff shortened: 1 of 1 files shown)


phimage:
I think the problem is variable naming, the parameters must not the one from callback

so you can test modifying 
`success: { (credential, response, parameters) in`
by
`success: { (credential, response, param) in` or `success: { (credential, _, _) in`
and close this PR if this work for you

boyvanamstel:
Alright, wasn't sure if the parameters needed to be passed along or not. I've updated the repo with a new commit.

This works as expected as well.

phimage:
thx

PR #311, opened 2016-11-12
Fix OAuthSwiftCredential.Version init(:) error

bzmario:


phimage:
Thx, so this fix storage with NSCoding

PR #325, opened 2016-12-21
Fix: encode redirect_uri

munkius:
The redirect_uri was not encoded. This causes issues when the redirect_uri contains query parameters. Encoding the uri makes sure the entire string is used, and not mixed up with and other parameters on the url.


phimage:
thanks
I will try to add a unit test for this

phimage:
it's seems to have some issue with this commit 
issue #339

I will revert it, and put maybe a boolean config

PR #326, opened 2016-12-21
Fix typo

ayunav:
"You can extend `OAuthWebViewController`..." instead of "You can extends `OAuthWebViewController`..."


phimage:
thx

PR #335, opened 2017-02-10
Make requestOAuthAccessToken open

wesley-dynamicowl:
We have an interesting use case with Microsoft's OAuth (Azure AD, Graph API, Outlook API, etc)

Oauth tokens are scoped to a single resource. However, Microsoft allows refresh tokens to be used to request additional tokens for other resources.  The easiest way to do this is to call requestOAuthAccessToken for each additional resource we intend to use with the additional resource parameters we need to specify.

Opening requestOAuthAccessToken makes it much simpler for us to request additional access tokens.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -167,8 +167,8 @@ open class OAuth2Swift: OAuthSwift {
     }
     
     @discardableResult
-    open func renewAccessToken(withRefreshToken refreshToken: String, headers: OAuthSwift.Headers? = nil, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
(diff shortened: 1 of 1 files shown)


phimage:
ok for the need
But maybe there is another solution by keeping original public method

 say me if it could match your need

adding ` , parameters: OAuthSwift.Parameters` into `renewAccessToken` open function

then all parameters client_id, client_secret, refresh_token, grant_type will be filled and your custom parameters will be added 
so you can override it but cannot remove one

wesley-dynamicowl:
That would probably work well. It'd prevent unexpected behavior if people don't provide all required oauth parameters and still allow for custom parameters to be added.

I'll make those changes and see what it looks like

phimage:
thx for the update, next time do a new PR (history of project will be more clean)

PR #351, opened 2017-03-14
Created a variable that we can handle if want or not animation transition

felipeflorencio:
I have the need to set to false navigation transition, because my type of navigation UX, but using the webview provided by lib i can not do this, only if i change the lib, but will cause problem beside i'm using cocoapods as manager.

So i changed the variable "animated" to one that now is public and provide by default the animation as "true", but now i have the chance to set to "false" if i need.

Hope get merged so i will can use at my project without need of change all the flow or something like that


phimage:
thanks

Issue #356, opened 2017-04-04
How can I get callback that authorization process has been finished?


hiragram:
### Description:

I am using my custom webview controller, and I want to dismiss the vc when authorization has succeeded.
I have looked into implementation of `SafariURLHandler` and found it uses `OAuthSwift.CallbackNotification.notificationName` but `CallbackNotification` is internal struct so I cannot use.
Are there any way to receive callback like that?

### OAuth Version:
- [ ] Version 1
- [ ] Version 2

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [x] Carthage
- [ ] CocoaPods
- [ ] Manually

### Library version:
- [ ] head
- [x] v1.0.0
- [ ] v0.6
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 8.0 (Swift 3.0)
- [ ] 8.0 (Swift 2.3)
- [ ] 7.3.1
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

phimage:
You can look at demo app
https://github.com/OAuthSwift/OAuthSwift/blob/0296377d14b27d89113afbf15fc0f01f7f06f304/Demo/Common/WebViewController.swift
Using the delegate of web view you could know when your url is open and so you can dismiss immediately or launch a task using DispatchQueue

hiragram:
@phimage 
Thanks for quick reply.
I understand your suggestion. But my webview controller is used in whole app so I don't want to write OAuth-specific code in the webview controller class. If I could receive notification in my custom `OAuthSwiftURLHandlerType` such as `SafariURLHandler` , that sounds more better to me.
How do you think about making public `CallbackNotification` type?

phimage:
why not, but the code is not clean

For instance I think instead of
```swift
 struct CallbackNotification {
        static let notificationName = Notification.Name(rawValue: "OAuthSwiftCallbackNotificationName")
    }
```
we must have

```swift 
extension Notification.Name {
    public static var OAuthSwiftHandleCallbackURL: Notification.Name = .init("OAuthSwiftCallbackNotificationName")
}
```

If you do the PR I will merge

hiragram:
@phimage 
Sounds cool. I will do that on my fork and create PR.

PR #357, opened 2017-04-04
Making notification name public

hiragram:
This PR is related to #356 .

I wanted to receive callback for authorization process was finished in my custom `OAuthSwiftURLHandlerType` .
But `CallbackNotification` is internal struct so I cannot use from outside of module.

For more detail, see conversation in #356 .


phimage:
you could close your issue ;)

PR #364, opened 2017-04-23
Fixed swift 3.1 compiler warnings

patricks:


phimage:
Thanks
I will make some modification

wattson12:
@phimage as these changes aren't in 1.1.1, would it be possible to get a version bump? thanks

phimage:
just a tag?

wattson12:
I'm using cocoapods so ideally a tag + update to the podspec, but can point to a tag as well if there is one

phimage:
so you do not need a tag or podspec push, just use the head version by specifying the git url (or point on specific commit)
I cannot do podspec push, I have no access to a computer for several days

wattson12:
OK no worries, thanks for the quick response (btw I'm happy to submit a PR with podspec changes for when it can be pushed, though would need to know the new version number)

PR #369, opened 2017-05-03
Fix a crash when encoding credentials in a test target

eunikolsky:
When `NSCoding` an `OAuthSwiftCredential` instance, the test would crash
on `CodingKeys` initialization because the `bundleIdentifier` of the
`Bundle.main` in the tests is `nil` (the main bundle is not app's
bundle). This patch fixes that by trying to get one of the bundle
identifiers: the `main` one, the bundle for the `OAuthSwiftCredential`
class, or an empty string otherwise; a test is added to verify the fix.


phimage:
thanks!!

PR #378, opened 2017-05-29
Adding convenience initializer and documentation for OAuth 1.0a requests

josephquigley:
The readme is unclear about how to make OAuth 1.0a requests. I added a convenience initializer that will set all the unnecessary `OAuth1Swift()` init params to the required empty strings to get it to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,22 @@ let handle = oauthswift.authorize(
     }             
 )
 ```
+
+### Request example with OAuth1.0a
+```swift
+// create an instance and retain it
+oauthswift = OAuth1Swift(
+consumerKey:    "********",
+consumerSecret: "********")
+
+oauthswift.client.get("http://api.example.com/foo/bar", success: { data in
+//....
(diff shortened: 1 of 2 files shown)


josephquigley:
I tested it with the [Noun Project API](http://api.thenounproject.com/getting_started.html#sample-code):
```
let oauthswift = OAuth1Swift(consumerKey: "********", consumerSecret: "********")
oauthswift.client.get("http://api.thenounproject.com/icon/apple", success: { data in 
   print(data)
}, failure: { error in
   print(error.localizedDescription
})
```
Regarding the authorize error I could create a class, `OAuth1aSwift` that conforms to `OAuthSwift` and doesn't have an authorize method since authorization callbacks are not used by the 1.0a protocol.

As for urls, not sure what to suggest there. I added the example of how to use 1.0a requests in the `README.md` so hopefully people will see that. If I had seen that I wouldn't have opened an issue ;).

PR #383, opened 2017-06-15
readme/atlassian

Noambaron:
Adds Jira's and updates BitBucket's OAuth provider pages. 
Also fixes a small typo.


phimage:
thanks for the fix

But adding Jira means adding also it into demo app to validate that there is no compatibility issue
https://github.com/OAuthSwift/OAuthSwift/wiki/Demo-application#add-a-new-service-in-demo-app

PR #390, opened 2017-07-06
Refactor o auth swift credential

Noambaron:
This refactors OAuthSwiftCredentials to be more flexible and extendable. No underlying implementation changed.
 
It uses a protocol based approach with base classes that implement common functionality, and concrete subclasses for specific credentials. This approach allows easy adjustment and enhancement and lays the ground for RSA support (That will follow in next PR). 

OAuthSwiftCredentials is broken down into 3 protocols (OAuthSwiftCredential, OAuthSwiftSignatureMethod and OAuthSwiftVersion), There are also 3 base classes respectfully implementing common aspects of the protocols, and concrete classes for OAuth1, OAuth2 and HMAC-SHA1 that implement the specific parts. 

Next PR will add RSA-SHA1 signatureMethod to extend functionality.

### DIFF
--- a/Demo/Common/Images.xcassets/AppIcon.appiconset/Contents.json
+++ b/Demo/Common/Images.xcassets/AppIcon.appiconset/Contents.json
@@ -91,6 +91,11 @@
       "idiom" : "ipad",
       "filename" : "oauthswift-icon-153.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ios-marketing",
+      "size" : "1024x1024",
+      "scale" : "1x"
     }
   ],
   "info" : {
--- a/Sources/Credential/Credential Base Classes/BaseOAuthSwiftCredential.swift
+++ b/Sources/Credential/Credential Base Classes/BaseOAuthSwiftCredential.swift
@@ -1,101 +1,44 @@
 //
-//  OAuthSwiftCredential.swift
+//  BaseOAuthSwiftCredential.swift
 //  OAuthSwift
 //
-//  Created by Dongri Jin on 6/22/14.
-//  Copyright (c) 2014 Dongri Jin. All rights reserved.
+//  Created by Noam Bar-on on 6/21/17.
+//  Copyright © 2017 Dongri Jin. All rights reserved.
 //
+
 import Foundation
 
-// Allow to customize computed headers
-public protocol OAuthSwiftCredentialHeadersFactory {
-    func make(_ url: URL, method: OAuthSwiftHTTPRequest.Method, parameters: OAuthSwift.Parameters, body: Data?) -> [String: String]
-}
-
-// The credential for authentification
-open class OAuthSwiftCredential: NSObject, NSCoding {
-
-    public enum Version {
-        case oauth1, oauth2
-
-        public var shortVersion: String {
-            switch self {
-            case .oauth1:
-                return "1.0"
-            case .oauth2:
-                return "2.0"
-            }
-        }
-
-        public var signatureMethod: SignatureMethod {
-            return .HMAC_SHA1
-        }
-
-        var toInt32: Int32 {
-            switch self {
-            case .oauth1:
-                return 1
-            case .oauth2:
-                return 2
-            }
-        }
-        init(_ value: Int32) {
-            switch value {
-            case 1:
-                self = .oauth1
-            case 2:
-                self = .oauth2
-            default:
-                self = .oauth1
-            }
-        }
-    }
-
-    public enum SignatureMethod: String {
-        case HMAC_SHA1 = "HMAC-SHA1"//, RSA_SHA1 = "RSA-SHA1", PLAINTEXT = "PLAINTEXT"
-
-        func sign(key: Data, message: Data) -> Data? {
-            switch self {
-            case .HMAC_SHA1:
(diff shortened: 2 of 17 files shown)


phimage:
Thanks for the contribution.

Support for RSA-SHA1 is welcome! 
But I don't think that all this change is needed or maybe I miss something about how to sign with that RSA-SHA1, If message or data are encoded or not in a same way 
because we can just uncomment `RSA_SHA1 = "RSA-SHA1` case from `SignatureMethod` and call the code inside the enum to encode

Sources/Credential/Credential Base Classes/BaseOAuthSwiftCredential.swift
```diff
@@ -119,9 +61,9 @@ open class OAuthSwiftCredential: NSObject, NSCoding {
         self.oauthTokenSecret = (decoder.decodeObject(forKey: CodingKeys.oauthTokenSecret) as? String) ?? String()
         self.oauthVerifier = (decoder.decodeObject(forKey: CodingKeys.oauthVerifier) as? String) ?? String()
         self.oauthTokenExpiresAt = (decoder.decodeObject(forKey: CodingKeys.oauthTokenExpiresAt) as? Date)
-        self.version = Version(decoder.decodeInt32(forKey: CodingKeys.version))
+        self.version = (decoder.decodeObject(forKey: CodingKeys.version) as! OAuthSwiftVersion)
     }
```
phimage: old stored credentials will not be readable (one way is to change the key and try to read the two)
force cast is not recommended, If you have swiftlint, there is an error

Sources/OAuthSwiftClient.swift
```diff
@@ -16,7 +16,7 @@ var OAuthSwiftDataEncoding: String.Encoding = .utf8
 
 open class OAuthSwiftClient: NSObject {
 
-    fileprivate(set) open var credential: OAuthSwiftCredential
+    open var credential: OAuthSwiftCredential
```
phimage: `fileprivate(set)` has been added to not allow someone to put a credential associated with a version 1 in a client for oauth 2 and vis versa
If you need to remove it, something must change to check that

Sources/OAuthSwiftClient.swift
```diff
@@ -30,19 +30,19 @@ open class OAuthSwiftClient: NSObject {
     public init(credential: OAuthSwiftCredential) {
         self.credential = credential
     }
-
-    public convenience init(consumerKey: String, consumerSecret: String, version: OAuthSwiftCredential.Version = .oauth1) {
-        let credential = OAuthSwiftCredential(consumerKey: consumerKey, consumerSecret: consumerSecret)
-        credential.version = version
+
```
phimage: this is a public init, so a breaking change
to maintain public api we can keep the old init, and call the new one. then add the deprecated annotation if the old one is no more necessary

Sources/Credential/Credential Base Classes/BaseOAuthSwiftCredential.swift
```diff
@@ -130,125 +72,103 @@ open class OAuthSwiftCredential: NSObject, NSCoding {
...
-            return self.oauthToken.isEmpty ? [:] : ["Authorization": "Bearer \(self.oauthToken)"]
-        }
+        preconditionFailure("This abstract method must be overridden")
     }
-
+    
     open func authorizationHeader(method: OAuthSwiftHTTPRequest.Method, url: URL, parameters: OAuthSwift.Parameters, body: Data? = nil) -> String {
         let timestamp = String(Int64(Date().timeIntervalSince1970))
-        let nonce = OAuthSwiftCredential.generateNonce()
+        let nonce = BaseOAuthSwiftCredential.generateNonce()
```
phimage: SignTests.swift compile?
generateNonce is public some same here there is a breaking change, but I do not see any solution

Demo/Common/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -91,6 +91,11 @@
       "idiom" : "ipad",
       "filename" : "oauthswift-icon-153.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ios-marketing",
+      "size" : "1024x1024",
+      "scale" : "1x"
     }
```
phimage: xcode 9 :)

Noambaron:
ok. I'm closing this PR. 
I opened a new PR instead of this one, with out refactoring and including the full code for RSA support and Jira example. 

as a side note: I still believe this refactoring (and new design) makes much more sense in terms of flexibility and encapsulation. I recognize it breaks current api, but I would still urge to consider it. I'd be happy to help work through the deprecation strategy.

phimage:
I am a former Java developper, I am agree with interfaces/protocols for flexibility
But as a swift developper I prefer to use clean swift feature, until a feature is impossible to implement in a clean way.

swift 4 version could be an api break, If we really need it.

Noambaron:
Protocols are at the [heart of Swift ](https://developer.apple.com/videos/play/wwdc2015/408/), so they are considered the most clean swift feature possible.

I agree that Swift 4 might be a good opportunity for api breaking changes. 
ping me on nbaron@atlassian.com when you get a chance - I'd love to chat some more, and I have a few questions for you

PR #394, opened 2017-07-12
RSA and Jira support

Noambaron:
This adds OAuth1 SHA1-RSA and Jira support

### DIFF
--- a/Demo/Common/ViewController.swift
+++ b/Demo/Common/ViewController.swift
@@ -131,6 +131,8 @@ extension ViewController {
             doOAuthSalesforce(parameters)
         case "BitBucket":
             doOAuthBitBucket(parameters)
+        case "Jira":
+            doOAuthJira(parameters)
         case "GoogleDrive":
             doOAuthGoogle(parameters)
         case "Smugmug":
@@ -685,7 +687,7 @@ extension ViewController {
             success: { credential, response, parameters in
                 self.showTokenAlert(name: serviceParameters["name"], credential: credential)
                 self.testBitBucket(oauthswift)
-            },
+        },
             failure: { error in
                 print(error.description)
             }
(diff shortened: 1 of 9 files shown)


Demo/Common/ViewController.swift
```diff
@@ -705,6 +707,42 @@ extension ViewController {
         )
     }
     
+    // MARK: Jira
+    func doOAuthJira(_ serviceParameters: [String:String]){
+        assert(Bundle.main.path(forResource: "private", ofType: "pem") != nil, "Jira authentication uses RSA-SHA1 which requires a stighlty different setup - see README.md for details.")
+        let oauthswift = OAuth1Swift.initWithRSA(
```
phimage: could add something like that 
```swift
let jiraURL = serviceParameters["url"] ?? "https://yourinstance.atlassian.net"
```
and then use it in the following code, allowing to use in config file a custom url

Sources/RSA.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  RSA.swift
+//  OAuthSwift
+//
+//  Created by Noam Bar-on on 6/19/17.
+//  Copyright © 2017 Dongri Jin. All rights reserved.
+//
+
+import Foundation
+import SwiftyRSA
```
phimage: So OAuthSwift become dependant on SwiftyRSA
So OAuthSwift.podspec and Package.swift must be updated

But for carthage a Cartfile must be created, and see the comment on Podfile

Podfile
```diff
@@ -1,5 +1,15 @@
 use_frameworks!
 
+target 'OAuthSwift' do
+    platform :ios, '8.3'
+    pod 'SwiftyRSA', '~> 1.2.0'
+end
+
```
phimage: By becoming dependant and make OAuthSwift target dependant on this I suspect that the project file has been updated to work only with cocoapod
So opening the projet without the workspace will failed
And then Carthage could not work.

I could do the fix later

Then for macOS? :p

Sources/OAuthSwiftCredential.swift
```diff
@@ -16,24 +16,24 @@ public protocol OAuthSwiftCredentialHeadersFactory {
 open class OAuthSwiftCredential: NSObject, NSCoding {
 
     public enum Version {
-        case oauth1, oauth2
+        case oauth1, oauth1RSA, oauth2
```
phimage: After merging I will do some change, try to remove this case oauth1RSA
by cutting the link between version and signature or use oauth1(SignatureMethod)

phimage:
I make some change in last version allowing to make dependencies injection.

The class `RSA` in this PR could implement the new protocol `OAuthSwiftSignatureDelegate`

then to inject 
```swift
OAuthSwiftCredential.SignatureMethod.delegates[.RSA_SHA1: RSA.Type)
```

So a new project `OAuthSwiftRSA` dependent on OAuthSwift and SwiftyRSA could be created
with the `RSA` class
A specific demo with JIRA could be created.

PR #397, opened 2017-07-25
Comment formatting

jrtibbetts:
This is a simple PR that corrects the formatting of class, function, and property comments so that they'll be parsed correctly and show up when option-clicking symbols. There are zero changes to functionality.


phimage:
Thanks

PR #402, opened 2017-08-15
Add encode callback url query option

### LINKED ISSUE #401: Double encoded callback url causes issues
### Description:

I've been using the framework successfully for awhile, but recently it stoped working with Imgur with `The redirect URI provided is missing or does not match` error. The issue doesn't seem to be related to any particular changes in the framework, but rather to Imgur backend updates. However, oauth works with Postman and Paw, which leads to certain conclusions.

Digging deeper the issue seems to be related to double encoded callback url in `OAuth2Swift::authorize` method. Removing double encoding solves the issue. I believe services take this into account and un-encode the url, which is good and bad, as it makes it easier to work with, but introduces uncertainty.

[`queryString` composed in `OAuth2Swift::authorize`](https://github.com/OAuthSwift/OAuthSwift/blob/b50109fd752f93c93151de4b7695bd710aa26bed/Sources/OAuth2Swift.swift#L121-L143) combines every parameter as is with exception for callback url, which gets encoded. The composed query then gets encoded, which causes url to be double encoded. If query would be composed with every parameter encoded immediately and without re-encoding everything, it would produce a valid query and eliminate the issue.

P. S. I see that logic [has changed recently](https://github.com/OAuthSwift/OAuthSwift/blob/b50109fd752f93c93151de4b7695bd710aa26bed/Sources/OAuth2Swift.swift#L122-L126), however, the problem occurs in the following `urlQueryEncoded` invocation.

### OAuth Provider (Twitter, Github, ..):
- [x] Imgur

### OAuth Version:
- [x] Version 2

- [x] OSX :

### Installation method:
- [x] Carthage

### Library version:
- [x] v1.1.2

### Xcode version:
- [x] 8.0 (Swift 3.0)

ianbytchek:
This gives control over double encoding and fixes #401.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -111,7 +111,7 @@ open class OAuth2Swift: OAuthSwift {
                 }
                 if let handle = this.postOAuthAccessTokenWithRequestToken(
                     byCode: code.safeStringByRemovingPercentEncoding,
-                    callbackURL: URL(string: callbackURL.absoluteString.urlEncodedString)!, headers: headers, success: success, failure: failure) {
+                    callbackURL: URL(string: callbackURL.absoluteString.urlEncoded)!, headers: headers, success: success, failure: failure) {
                     this.putHandle(handle, withKey: UUID().uuidString)
                 }
             } else if let error = responseParameters["error"] {
@@ -127,7 +127,7 @@ open class OAuth2Swift: OAuthSwift {
         var queryString: String? = ""
 
         queryString = queryString?.urlQueryByAppending(parameter: "client_id", value: self.consumerKey)
-        queryString = queryString?.urlQueryByAppending(parameter: "redirect_uri", value: self.encodeCallbackURL ? callbackURL.absoluteString.urlEncodedString : callbackURL.absoluteString, encode: self.encodeCallbackURLQuery)
+        queryString = queryString?.urlQueryByAppending(parameter: "redirect_uri", value: self.encodeCallbackURL ? callbackURL.absoluteString.urlEncoded : callbackURL.absoluteString, encode: self.encodeCallbackURLQuery)
         queryString = queryString?.urlQueryByAppending(parameter: "response_type", value: self.responseType)
         queryString = queryString?.urlQueryByAppending(parameter: "scope", value: scope)
         queryString = queryString?.urlQueryByAppending(parameter: "state", value: state)
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -17,9 +17,12 @@ open class OAuth2Swift: OAuthSwift {
     /// Set to true to deactivate state check. Be careful of CSRF
     open var allowMissingStateCheck: Bool = false
 
-    /// Encode callback url. Default false
-    /// issue #339, pr ##325
+    /// Encode callback url, some services require it to be encoded.
     open var encodeCallbackURL: Bool = false
+
+    /// Encode callback url inside the query, this is second encoding phase when the entire query string gets assembled. In rare 
+    /// cases, like with Imgur, the url needs to be encoded only once and this value needs to be set to `false`.
+    open var encodeCallbackURLQuery: Bool = true
 
     var consumerKey: String
     var consumerSecret: String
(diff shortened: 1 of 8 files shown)


Sources/OAuth2Swift.swift
```diff
@@ -118,32 +124,24 @@ open class OAuth2Swift: OAuthSwift {
...
-            queryString += "&\(param.0)=\(param.1)"
-        }
+        var queryString: String? = ""
 
-        var urlString = self.authorizeUrl
-        urlString += (self.authorizeUrl.contains("?") ? "&" : "?")
+        queryString = queryString?.urlQueryByAppending(parameter: "client_id", value: self.consumerKey)
+        queryString = queryString?.urlQueryByAppending(parameter: "redirect_uri", value: self.encodeCallbackURL ? callbackURL.absoluteString.urlEncoded : callbackURL.absoluteString, encode: self.encodeCallbackURLQuery)
+        queryString = queryString?.urlQueryByAppending(parameter: "response_type", value: self.responseType)
+        queryString = queryString?.urlQueryByAppending(parameter: "scope", value: scope)
```
phimage: some providers do not support empty parameters like scope, parameter, that`s why original code have `if !scope.isEmpty {` ` if !state.isEmpty {`

Sources/OAuth2Swift.swift
```diff
@@ -290,3 +289,13 @@ open class OAuth2Swift: OAuthSwift {
     }
 
 }
+
+extension OAuth2Swift {
+
+    public enum CallbackURLEncode {
```
phimage: you choose the two boolean option but keep an unused enum

Sources/OAuth2Swift.swift
```diff
@@ -118,32 +124,24 @@ open class OAuth2Swift: OAuthSwift {
...
+            queryString = queryString?.urlQueryByAppending(parameter: name, value: "\(value)")
+        }
+
+        if let queryString: String = queryString, let url: URL = URL(string: self.authorizeUrl.urlByAppending(query: queryString)) {
+            self.authorizeURLHandler.handle(url)
             return self
         } else {
             self.cancel() // ie. remove the observer.
-            failure?(OAuthSwiftError.encodingError(urlString: urlString))
+            failure?(OAuthSwiftError.encodingError(urlString: self.authorizeUrl))
```
phimage: here we loose information about what could not be encoded/converted to URL
maybe replace by `self.authorizeUrl.urlByAppending(query: queryString`

Sources/OAuth2Swift.swift
```diff
@@ -118,32 +124,24 @@ open class OAuth2Swift: OAuthSwift {
...
-        if !scope.isEmpty {
-            queryString += "&scope=\(scope)"
-        }
-        if !state.isEmpty {
-            queryString += "&state=\(state)"
-        }
-        for param in parameters {
-            queryString += "&\(param.0)=\(param.1)"
-        }
+        var queryString: String? = ""
```
phimage: optional?

ianbytchek: To make this work, parameters are encoded as they are added, which may produce `nil` value, just like when you [encode the whole string](https://github.com/OAuthSwift/OAuthSwift/blob/b50109fd752f93c93151de4b7695bd710aa26bed/Sources/OAuth2Swift.swift#L141). Here we add them and if everything is ok, we end up with the same encoded query string, if not – it's `nil`, which we [check in the same fashion](https://github.com/ianbytchek/OAuthSwift/blob/3915fcace18ede9bf21d7767a16ac80df967db96/Sources/OAuth2Swift.swift#L139) and [fail with the same error](https://github.com/ianbytchek/OAuthSwift/blob/3915fcace18ede9bf21d7767a16ac80df967db96/Sources/OAuth2Swift.swift#L144).

Sources/OAuth2Swift.swift
ianbytchek: That's fine, they are [checked in `urlQueryByAppending` and not added if empty](https://github.com/ianbytchek/OAuthSwift/blob/3915fcace18ede9bf21d7767a16ac80df967db96/Sources/String%2BOAuthSwift.swift#L30).

Sources/OAuth2Swift.swift
ianbytchek: I don't think we do – [`urlString` is defined a couple lines before](https://github.com/OAuthSwift/OAuthSwift/blob/b50109fd752f93c93151de4b7695bd710aa26bed/Sources/OAuth2Swift.swift#L138) and only gets appended with `?` or `&` in order to make it ready for [concatenation with with query string below](https://github.com/OAuthSwift/OAuthSwift/blob/b50109fd752f93c93151de4b7695bd710aa26bed/Sources/OAuth2Swift.swift#L141). When failure gets invoked, the variable would only differ from the property by one extra character.

Sources/OAuth2Swift.swift
ianbytchek: Ooops, forgot about it. I've tried using enum as suggested, but realised that naming might be an issue and would cause issues for users, who already use the initial property, so went back to boolean forgetting to remove the declaration. Well noticed.

ianbytchek: Fixed.

Sources/OAuth2Swift.swift
phimage: yes, the previous code is also wrong (because of encodedQuery, same here if no queryString )

Sources/OAuth2Swift.swift
phimage: ok

Sources/OAuth2Swift.swift
phimage: ok
so there is a behaviour change for response_type , client_id, redirect_uri
but I don't think that this could add issue

phimage:
thanks for the PR
I will do a test and merge when I have some times

Sources/OAuth2Swift.swift
ianbytchek: I thought that it might be nice to use [`URLComponents`](https://developer.apple.com/documentation/foundation/urlcomponents) approach with custom encoding, but thought that it might be a radical addition for this PR. Using string extensions and doing low-level url assembly in authorisation method feels wrong. I'm happy to make another PR if you are happy with the approach.

Issue #400, opened 2017-08-14
Retrieving only authentication code by specifying response_type as code


uiroshan:
### Description:
I am trying to implement a login flow, which involves both iOS app and server integration. Other than social authentication, application support email/password authentication as well. Once successful authentication server application send auth_token which can be used in subsequent API calls.

My social login flow as follows : 
* iOS application initiate social login by retrieving a code as response type from the authentication provider. 
* Retrieved code send to server application, so it verifies with the social authentication provider and (create an account if not exist) authenticate the user. 
* Successful authentication will generate auth token as normal (email / password) authentication and send back to the mobile application. 

Once successful authentication, both approaches mobile application receives auth_token.  

But the problem with OAuthSwift is, it doesn't support retrieving just the code from social provider. Once we specify response type as code, it needs consumerSecret and accessTokenUrl as required parameters and completes access_token retrieving step in the background. Which I am expected to do as a server-server call. Also Facebook recommends to exchange code for the access_token as a server-server call. 

Is there any provided way to do this?

### OAuth Provider (Twitter, Github, ..):
Facebook

### OAuth Version:
- [ ] Version 1
- [x] Version 2

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [x] CocoaPods
- [ ] Manually

### Library version:
- [x] head
- [ ] v1.0.0
- [ ] v0.6
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 8.3.3 (Swift 3.0)
- [ ] 8.0 (Swift 3.0)
- [ ] 8.0 (Swift 2.3)
- [ ] 7.3.1
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

phimage:
If I understand well your workflow you want to split the current workflow

In OAuthSwift2 `authorize`method
we receive the code `let code = responseParameters["code"]`
and make a new request using `this.postOAuthAccessTokenWithRequestToken``

So maybe you want to override `postOAuthAccessTokenWithRequestToken(byCode:`

- So one way is to make this function accessible and overridable then in your code you can extend OAuthSwift2 to fulfil your need.

- The other way is to add a new attribute in OAuthSwift2, a closure (or a protocol like a delegate).
 If the closure is provided, then call the closure code instead of `postOAuthAccessTokenWithRequestToken(byCode:

[...]

PR #403, opened 2017-08-18
Retrieving only authentication code by specifying response_type as code

uiroshan:
Fixes for enhancement #400

I made it open and changed the base class name to exchangeCodeForAccessToken. I thought method needs a better name so use the new name. But if you think it's not necessary or does not match with your naming conventions please let me know.

Gist link on how to use
https://gist.github.com/uiroshan/89f9baea0a2678955bc672bfcefbfebd


phimage:
Thanks for the PR
I prefer to keep the current name
I put public

Issue #407, opened 2017-09-08
Carthage build failing due to SwiftLint


col:
### Description:

Run ``carthage build``

Output

```
The following build commands failed:
	PhaseScriptExecution swiftlint /Users/col/Library/Developer/Xcode/DerivedData/OAuthSwift-dnqdsjigmwdntdbelfgdypovmokj/Build/Intermediates/OAuthSwift.build/Release-iphoneos/OAuthSwift.build/Script-C41824251EE4079C00C5D0A3.sh
(1 failure)
ift/Sources/String+OAuthSwift.swift:88:40: error: Superfluous Disable Command Violation: SwiftLint rule 'force_cast' did not trigger a violation in the disabled region. Please remove the disable command. (superfluous_disable_command)
```

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [x] Carthage
- [ ] CocoaPods
- [ ] Manually

### Library version:
- [x] head
- [x] v1.1.2
- [ ] v0.6
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 8.3.3 (Swift 3.1)
- [ ] 8.0 (Swift 3.0)
- [ ] 8.0 (Swift 2.3)
- [ ] 7.3.1
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

phimage:
Thanks for the issue
I cannot fix with #408 because not everybody update swiftlint 
And I think swiftlint is buggy for that rule...

I will fix by setting Superfluous Disable Command Violation  as warning

phimage:
I try to fix my push but I have an issue with git
```
git -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree push -v --tags https://github.com/OAuthSwift/OAuthSwift refs/heads/master:refs/heads/master
Pushing to https://github.com/OAuthSwift/OAuthSwift
remote: Permission to OAuthSwift/OAuthSwift.git denied to phimage.
fatal: unable to access 'https://github.com/OAuthSwift/OAuthSwift/': The requested URL returned error: 403
Completed with errors, see above
```

PR #408, opened 2017-09-08
Remove unnecessary swiftlint comment

col:
It's not necessary to disable force_cast here as the code is not causing a force_cast error.

This is actually causing my Carthage build to fail. :cry:

See Issue: [#407](https://github.com/OAuthSwift/OAuthSwift/issues/407)


phimage:
see comment on issue

PR #411, opened 2017-09-24
Added comment regarding retaining instance

danjohnson95:
Hey there,

Today I got stuck with a problem regarding the `-10 retain` error.

After checking the issues, I noticed that [11 people](https://github.com/OAuthSwift/OAuthSwift/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20retain) also got stuck with that same problem.

When initially reading the docs, I skipped straight past OAuth1 and just looked at the code snippet for OAuth2. Maybe adding the comment about retention in this code snippet will help others, and hopefully will reduce the number of issues raised for the same problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ oauthswift.client.get("https://api.example.com/foo/bar",
 
 ### Authorize with OAuth2.0
 ```swift
+// create an instance and retain it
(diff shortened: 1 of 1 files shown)


phimage:
Thanks
Good idea

PR #438, opened 2018-02-26
Fix incorrect coding key for signatureMethod on OAuthSwiftCredential when using a version of .oAuth1

Michael-Place:
The encode method was writing the value for `signatureMethod` to the key that was reserved for version. There was a type mismatch of `String` and `Int32` which causes a crash on decode.


phimage:
Thanks for the fix

PR #456, opened 2018-04-20
Add authorization by username and password

antondevs:
Please add authorization by username and password

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -289,14 +289,19 @@ open class OAuth2Swift: OAuthSwift {
     }
 	
 	// OAuth 2.0 Specification: https://tools.ietf.org/html/draft-ietf-oauth-v2-13#section-4.3
-	@discardableResult
+    @discardableResult
     open func authorize(username: String, password:String, scope:String?, headers: OAuthSwift.Headers? = nil, success: @escaping TokenSuccessHandler, failure: @escaping OAuthSwiftHTTPRequest.FailureHandler) -> OAuthSwiftRequestHandle? {
+    
         var parameters = OAuthSwift.Parameters()
         parameters["client_id"] = self.consumerKey
         parameters["client_secret"] = self.consumerSecret
         parameters["username"] = username
         parameters["password"] = password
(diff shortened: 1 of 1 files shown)


phimage:
Thanks
@antondevs did you have some exemple of service/provider that use this authorization mode?
Better if I can create a free account

Then the service could be added to the demo

Sources/OAuth2Swift.swift
```diff
@@ -287,6 +287,25 @@ open class OAuth2Swift: OAuthSwift {
...
+    open func authorize(username: String, password:String, success: @escaping TokenRenewedHandler, failure: @escaping OAuthSwiftHTTPRequest.FailureHandler) -> OAuthSwiftRequestHandle? {
+        var parameters = OAuthSwift.Parameters()
+        parameters["client_id"] = self.consumerKey
+        parameters["client_secret"] = self.consumerSecret
+        parameters["username"] = username
+        parameters["password"] = password
+        parameters["grant_type"] = "password"
+    
+        print(parameters)
+
```
phimage: no print in code

Sources/OAuth2Swift.swift
```diff
@@ -287,6 +287,25 @@ open class OAuth2Swift: OAuthSwift {
             }
         }
     }
+	
+	@discardableResult
+    open func authorize(username: String, password:String, success: @escaping TokenRenewedHandler, failure: @escaping OAuthSwiftHTTPRequest.FailureHandler) -> OAuthSwiftRequestHandle? {
+        var parameters = OAuthSwift.Parameters()
```
phimage: why `TokenRenewedHandler` ? copied from renew function
maybe SuccessHandler

phimage:
please put in function comment the link about the rfc

then could you add also optional `scope: String?` as the rfc describe 
and also `headers: OAuthSwift.Headers? = nil`

antondevs:
Updated. I added headers and comment. Scope param is not exist in requestOAuthAccessToken. Please apply

antondevs:
Please review. I am not using scope, also added scope param to func

PR #458, opened 2018-04-24
redirect_uri is optional in RFC 6749

slizeray:
callbackURL parameter is optional because redirect_uri parameter is optional in RFC 6749


phimage:
Please put a link to the rfc chapter about that

slizeray:
@phimage Could someone review this PR? Thanks ahead.

PR #461, opened 2018-05-15
Fix failing oauth1 tests Fix #460

### LINKED ISSUE #460: OAuthSwift1 tests are failing
### Description:
The two tests that are testing to make sure failures don't call their success handlers are failing because they're calling their success handlers. This is happening because the server they're talking to is returning a string and a 200 and the `authorize` method considers that a success without making sure there's a non-empty token value provided in the response.

### OAuth Version:
- [x] Version 1
- [ ] Version 2

- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [ ] CocoaPods
- [ ] Manually

### Library version:
- [x] head
- [ ] v1.2.1
- [ ] v1.2 (Swift 4.0)
- [ ] v1.0.0
- [ ] v0.6
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 9.3 (Swift 4.1)
- [ ] 9.0 (Swift 4.0)
- [ ] 9.0 (Swift 3.2)
- [ ] 8.x (Swift 3.x)
- [ ] 8.0 (Swift 2.3)
- [ ] 7.3.1
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

aranasaurus:
oauthbin is returning a 200 response with string data that explains the failure rather than a 400. The authorize method in this case is just using the OAuthSwiftCredential's default token value, which is an empty string. This change makes a response with an empty token value report as a `missingToken` error.

Fixes #460

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
     - PROJECT=OAuthSwift.xcodeproj
     - IOS_FRAMEWORK_SCHEME="OAuthSwift"
     - MACOS_FRAMEWORK_SCHEME="OAuthSwiftOSX"
-    - IOS_SDK=iphonesimulator11.0
+    - IOS_SDK=iphonesimulator11.3
     - MACOS_SDK=macosx10.13
   matrix:
     - DESTINATION="OS=11.0,name=iPhone 8" SCHEME="$IOS_FRAMEWORK_SCHEME" SDK="$IOS_SDK"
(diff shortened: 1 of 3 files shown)


aranasaurus:
those last two commits are just me trying to get CI to work. I don't know Danger, but it seems like that's where the problem is now?

I'm happy to change this PR (and my other one #463) with some guidance with regards to any of that.

phimage:
thanks for the PR

I start to add danger to ci file but not finish I think
So no issue with the PR

PR #471, opened 2018-06-19
Add startAuthorizedRequest optional body parameter to support POST

ynechaev:
This PR introduces changes needed to perform `POST` request using convenience function `startAuthorizedRequest` with HTTP body parameter.

Before this change self.client.request was performed without optional `body` parameter.


ynechaev:
Looks like there is danger<--TravisCI issue `For your GitHub repo, you need to expose: DANGER_GITHUB_API_TOKEN` causing build to fail.

PR #494, opened 2018-10-26
Updated to Swift 4.2 and made the Xcode-recommended project changes

jrtibbetts:
Using Xcode 10.0, I let it update the code to be Swift 4.2-compatible, and I made the project file changes that Xcode recommended.


phimage:
thanks you for the modification
As you can see CI failed

Could you update .travis.yml https://github.com/OAuthSwift/OAuthSwift/blob/master/.travis.yml
by setting xcode10

PR #511, opened 2019-02-06
URL type support

armansharvel:
Most of the public methods of this library accept string-typed `URL`s. This is not very convenient for people who prefer constructing URLs with `URL.appendingPathComponent()`. This method allows you not to worry about a missed or overwritten slash and keeps it very clean and obvious. 

In this PR I didn't change any part of the internal logic. Only duplicated public/open methods that accept string-typed `URL`s and changed the arguments to url-typed `URL`s that would call the original method.

### DIFF
--- a/Sources/OAuth1Swift.swift
+++ b/Sources/OAuth1Swift.swift
@@ -117,12 +117,12 @@ open class OAuth1Swift: OAuthSwift {
     }
 
     @discardableResult
-    open func authorize(withCallbackURL URL: URLConvertible, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
-        guard let url = URL.url else {
-              failure?(OAuthSwiftError.encodingError(urlString: URL.string))
+    open func authorize(withCallbackURL url: URLConvertible, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
+        guard let callbackURL = url.url else {
+              failure?(OAuthSwiftError.encodingError(urlString: url.string))
             return nil
         }
-        return authorize(withCallbackURL: url, success: success, failure: failure)
+        return authorize(withCallbackURL: callbackURL, success: success, failure: failure)
     }
 
     // 1. Request token
--- a/Sources/OAuth1Swift.swift
+++ b/Sources/OAuth1Swift.swift
@@ -24,6 +24,10 @@ open class OAuth1Swift: OAuthSwift {
     var accessTokenUrl: String
 
     // MARK: init
+	public convenience init(consumerKey: String, consumerSecret: String, requestTokenUrl: URL, authorizeUrl: URL, accessTokenUrl: URL) {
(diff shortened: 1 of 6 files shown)


phimage:
Thanks for the PR.

 It seems to have some issue with code  formatting

I think that having only URL internally would be better and like in Alamofire have URLConvertible to protocol to not multiply methods. String and URL implement it, url request too. (Maybe for a  swift5 version)

Here if we keep String internally, maybe a protocol could also help to keep string as string by returning self, and url to string using absoluteString

armansharvel:
Hi Eric, thanks for your feedback! 

Actually, I'm not sure what's wrong with the formatting.

I agree, that a protocol like this would be a much cleaner solution for the time being. Please, have a look at my implementation when you have time.

armansharvel:
Weird, it says `use of undeclared type 'URLConvertible'` but clearly it's recognizable locally in the Xcode project.

phimage:
> Weird, it says `use of undeclared type 'URLConvertible'` but clearly it's recognizable locally in the Xcode project.

You must add the file for all targets, one by OS
screenshot 2019-02-12 at 09 14 36

armansharvel:
Ah, of course! Thank you, will fix it later today.

Sources/OAuth1Swift.swift
```diff
@@ -117,9 +117,9 @@ open class OAuth1Swift: OAuthSwift {
     }
 
     @discardableResult
-    open func authorize(withCallbackURL urlString: String, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
-        guard let url = URL(string: urlString) else {
-              failure?(OAuthSwiftError.encodingError(urlString: urlString))
+    open func authorize(withCallbackURL URL: URLConvertible, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
```
phimage: Could you edit and use `url` instead of `URL` here and for other functions
Thanks

Sources/OAuth2Swift.swift
```diff
@@ -169,9 +169,9 @@ open class OAuth2Swift: OAuthSwift {
     }
 
     @discardableResult
-    open func authorize(withCallbackURL urlString: String, scope: String, state: String, parameters: Parameters = [:], headers: OAuthSwift.Headers? = nil, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
-        guard let url = URL(string: urlString) else {
-            failure?(OAuthSwiftError.encodingError(urlString: urlString))
+    open func authorize(withCallbackURL URL: URLConvertible, scope: String, state: String, parameters: Parameters = [:], headers: OAuthSwift.Headers? = nil, success: @escaping TokenSuccessHandler, failure: FailureHandler?) -> OAuthSwiftRequestHandle? {
```
phimage: so here too

phimage:
Merged
Thanks!!

armansharvel:
@phimage Thank you! But did the build pass CI checks? I still see a red cross in front of my commits

phimage:
I have a miss configuration with danger I think

main branch is ok after merging
[Build Status](https://travis-ci.org/OAuthSwift/OAuthSwift)

armansharvel:
Awesome! Thank you!

PR #517, opened 2019-02-21
Add optional headers to - postOAuthAccessTokenWithRequestToken

csauvage:
I've added the optional headers to the postOAuthAccessTokenWithRequestToken method.


csauvage:
@phimage any idea why Travis failed ? Seems that you don't expose Danger API keys

phimage:
the danger api key is declared but...

PR #538, opened 2019-08-13
Added support for macCatalyst

nrivard:
I added support for UIKitForMac (renamed `macCatalyst` in beta 4 and beyond). In order for this to work, I also had to fix `ASWebAuthenticationURLHandler` as it was not actually publicly usable. macCatalyst forbids the use of any deprecated functions so this had to be fixed. 

Tested on macOS Catalina beta 4 and 5.

### DIFF
--- a/Sources/OAuthSwiftURLHandlerType.swift
+++ b/Sources/OAuthSwiftURLHandlerType.swift
@@ -45,13 +45,16 @@ import SafariServices
 import AuthenticationServices
 #endif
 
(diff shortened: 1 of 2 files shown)


phimage:
thanks for the PR

because this change the method signature I cannot merge into master
(the only way is to have two version or make compatible code with default value for old version)

I will make a beta branch, with travis and xcode 11 too, and that into it

phimage:
I merged it into 2.1.0 branch with Travis running with Xcode 11
and project updated

PR #548, opened 2019-09-24
Update README.md for iOS 13

nonegu:
Before iOS 13, If a URL arrives while your app is suspended or running in the background, the system moves your app to the foreground prior to calling `application(_:open:options:)`. After Apple introduced `UISceneDelegate`, now system calls `scene(_:openURLContexts:)` when a URL arrives.

The readme section of **Handle URL in AppDelegate** revised accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,19 @@ func application(_ app: UIApplication, open url: URL, options: [UIApplication.Op
   return true
 }
 ```
+- On iOS 13, UIKit will notify `UISceneDelegate` instead of `UIApplicationDelegate`.
+- Implement `UISceneDelegate` method
+```swift
+func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+        guard let url = URLContexts.first?.url else {
+            return
+        }
+        if (url.host == "oauth-callback") {
+            OAuthSwift.handle(url: url)
+        }
+}
+```
+
 :warning: Any other application may try to open a URL with your url scheme. So you can check the source application, for instance for safari controller :
 ```
 if (options[.sourceApplication] as? String == "com.apple.SafariViewService") {


phimage:
thanks!!

I am wondering if we remove the scene configuration from AppDelegate, it continue to call the old method

```
    // MARK: UISceneSession Lifecycle

    func application(_ application: UIApplication, configurationForConnecting connectingSceneSession: UISceneSession, options: UIScene.ConnectionOptions) -> UISceneConfiguration {
        // Called when a new scene session is being created.
        // Use this method to select a configuration to create the new scene with.
        return UISceneConfiguration(name: "Default Configuration", sessionRole: connectingSceneSession.role)
    }

    func application(_ application: UIApplication, didDiscardSceneSessions sceneSessions: Set<UISceneSession>) {
        // Called when the user discards a scene session.
        // If any sessions were discarded while the application was not running, this will be called shortly after application:didFinishLaunchingWithOptions.
        // Use this method to release any resources that were specific to the discarded scenes, as they will not return.
    }
```

nonegu:
Removing UIScene configuration from `AppDelegate` does not work, it just crashes with an error of 'UISceneDelegate must be implemented'. Probably there are more links to remove in order to make `AppDelegate` responsible again.

But, according to the document on [developer.apple](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background), after `UISceneDelegate `successfully removed, `AppDelegate` should receive the notifications automatically. You may find the related section below:

> All state transitions result in UIKit sending notifications to the appropriate delegate object:
> In iOS 13 and later—A UISceneDelegate object.
> In iOS 12 and earlier—The UIApplicationDelegate object.
> You can support both types of delegate objects, but UIKit always uses scene delegate objects when they are available.

phimage:
when removing the scene delegate code we must add
```swift
    var window: UIWindow?
```
to come back to previous behaviour without scene

I also remove information about scene to info.plist to test it on new project
and it work with `func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {`

so if someone want to avoid scene I think it is possible (tested only on simulator)
so a previous created app will have no issue

slizeray:
@phimage In theory, you need to implement this method as well.....

    ```
func scene(_ scene: UIScene,
               willConnectTo session: UISceneSession,
               options connectionOptions: UIScene.ConnectionOptions) {
if let urlContext = connectionOptions.urlContexts.first {
        if (url.host == "oauth-callback") {
            OAuthSwift.handle(url: url)
            return
        }
        }}
```

nonegu:
> when removing the scene delegate code we must add
> 
> ```swift
>     var window: UIWindow?
> ```
> 
> to come back to previous behaviour without scene
> 
> I also remove information about scene to info.plist to test it on new project
> and it work with `func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {`
> 
> so if someone want to avoid scene I think it is possible (tested only on simulator)
> so a previous created app will have no issue

Oh, you are right. I did not remove the scene information from the info.plist, that was the reason why Xcode forces me to implement `UISceneDelegate`.

On the other hand, if the project created with previous versions of Xcode 11, then Xcode does not force you to use `UISceneDelegate` while migrating, so OAuthSwift still works as a charm for previously created projects! The issue is related with creating a new project with Xcode 11.

PR #558, opened 2019-11-15
fire dismiss handler when safariViewControllerDidFinish(control…

makanai5610:
I always use it conveniently. Thank you very much.

The dismiss completion did not work when the [DONE] button in Safari was tapped.
Since I had activated of UIActivityIndicatorView before transitioning to Safari, I needed dismissHandler to stop it.

I modified to work the dismissHandler when safariViewControllerDidFinish(:controller) fires.

Thank you for the review.

### DIFF
--- a/Sources/Handler/SafariURLHandler.swift
+++ b/Sources/Handler/SafariURLHandler.swift
@@ -114,6 +114,7 @@ open class SafariURLHandler: NSObject, OAuthSwiftURLHandlerType, SFSafariViewCon
     public func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
         // "Done" pressed
         self.clearObservers()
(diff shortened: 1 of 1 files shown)


phimage:
As you see in the code https://github.com/OAuthSwift/OAuthSwift/pull/558/files#diff-7e098e0e7e63844cfe77523492b6e00eR118 there is a delegate
If you have a special case because of your activity indicator why do not doing in this delegate
because I cannot merge PR if this is not a general behaviour for everyone (that's why I has added the delegate)

You can set it here
https://github.com/OAuthSwift/OAuthSwift/blob/63a4fdac45f7c016814729b49a5b129c1118f059/Sources/Handler/SafariURLHandler.swift#L34

PR #564, opened 2019-12-13
Includes all source subdirectories in pod spec with the recursive not…

slizeray:
@phimage Hello, I did a new pull request according to the changes requested. The podpsec now uses the recursive notation, and this is the single commit of the PR. Please let me know if there is any other issue. Thanks


phimage:
thanks

PR #565, opened 2019-12-13
Don't send client secret with PKCE

sprynmr:
Using Microsoft services I was getting an error when sending the client secret for a PKCE call.


sprynmr:
Looks like this is a CI issue, not with the code. Any maintainer around to check it out?

mesopelagique:
yes its a CI issue
thanks

phimage:
yes its a CI issue
thanks

PR #568, opened 2020-01-23
Add support for expired Twitter tokens

deyton:
See https://developer.twitter.com/en/docs/basics/response-codes

Twitter sends back errors in an errors array in the body, but it doesn't use any of the currently-supported method of indicating an expired or invalid token.


phimage:
thanks

PR #572, opened 2020-02-06
Fix new warnings when building with Xcode 11.4

deyton:
Xcode 11.4/Swift 5.2 came with a lot of changes in diagnostics, and there were a couple of warnings that popped up. I don't think there's anything risky here so hopefully it's helpful 😀

### DIFF
--- a/Sources/HMAC.swift
+++ b/Sources/HMAC.swift
@@ -35,9 +35,12 @@ open class HMAC {
 
         let ipadAndMessageHash = SHA1(ipad + message).calculate()
         let mac = SHA1(opad + ipadAndMessageHash).calculate()
-
-        return Data(bytes: UnsafePointer<UInt8>(mac), count: mac.count)
-
+        var hashedData: Data? = nil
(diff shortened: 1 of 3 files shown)


Sources/HMAC.swift
```diff
@@ -35,9 +35,12 @@ open class HMAC {
 
         let ipadAndMessageHash = SHA1(ipad + message).calculate()
         let mac = SHA1(opad + ipadAndMessageHash).calculate()
-
-        return Data(bytes: UnsafePointer<UInt8>(mac), count: mac.count)
-
+        var hashedData: Data? = nil
```
mesopelagique: this is art, transform one line code into closure...
`Data(withUnsafeBufferPointer: mac)` that done the same thing could be writted... (but yes maybe we can to do something else...

phimage: this is art, transform one line code into closure...
Data(withUnsafeBufferPointer: mac) that done the same thing could be writted... (but yes maybe we can to do something else...

phimage:
thanks

PR #576, opened 2020-03-05
If the client_secret is empty, don't post it at all.

sprynmr:
Really we [shouldn't be storing a client secret in an app at all](https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/), we should be using
PKCE. Some servers will reject the request if it's a "public" app and you include
the secret. Since it would be tricky to fully make it optional in the client everywhere,
I'm changing it to just ignore empty strings.

I didn't do this everywhere, as the other methods seem to assume the client/secret credentials are being used instead of pkce.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -181,11 +181,14 @@ open class OAuth2Swift: OAuthSwift {
         // PKCE - extra parameter
         if let codeVerifier = self.codeVerifier {
             parameters["code_verifier"] = codeVerifier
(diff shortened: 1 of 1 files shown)


sprynmr:
Any interest in pulling this in?

phimage:
Because of conflict due to other PR I will integrate the code
thanks
But I change a little, keep the else to not change the previous behaviour

majid701:
@phimage Can you publish a release that includes this? Looks like the latest release does not contain this change.

PR #578, opened 2020-03-16
Added support for watchOS < 6.0

vknabel:
Trying to build OAuthSwift for watchOS versions below 6.0 fail.
Adding the availability check fixes this problem.


phimage:
thanks!

PR #579, opened 2020-03-17
Don't send duplicate authentication with basic authentication

jimmya:
When enabling 
> accessTokenBasicAuthentification

The client_id and client_secret are still sent in the POST body. Some oAuth implementations complain about this. For example [Ory Hydra](https://github.com/ory/hydra) returns this error:
```
{
	"error": "invalid_client",
	"error_description": "Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method)",
	"error_hint": "The OAuth 2.0 Client supports client authentication method \"client_secret_basic\", but method \"client_secret_post\" was requested. You must configure the OAuth 2.0 client's \"token_endpoint_auth_method\" value to accept \"client_secret_post\".",
	"status_code": 401
}
```

This PR ensures that when authenticating through basic authentication, client_id and client_secret aren't sent in the post body.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -257,15 +257,18 @@ open class OAuth2Swift: OAuthSwift {
             return self.client.postMultiPartRequest(accessTokenUrl, method: .POST, parameters: parameters, headers: headers, checkTokenExpiration: false, completionHandler: completionHandler)
         } else {
             // special headers
+            var finalParameters = parameters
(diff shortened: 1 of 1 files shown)


phimage:
I cannot merge this PR. This change the behaviour for all.
Maybe some service need that. I have not time to test it all.
A configuration bool could do the trick to make it optional

FIndustries:
Spotify has the opposite issue: while using PKCE you must disable `accessTokenBasicAuthentification` but it fails if `client_secret` is set.

PR #583, opened 2020-03-31
Remove redundant URL initializer call from README

MaxDesiatov:
All these function calls take `URLConvertible` as an argument, and `String` is also `URLConvertible`. Thus `URL(string:)` initializer call is redundant and has a negative impact on readability of the example code in `README.md`.


phimage:
thanks

PR #209, opened 2016-03-28
[OAuth 2] Start an authorized request by suppling an automatic mechanism of token renewal

fabiomassimo:
# Summary

As described in OAuth 2 documentation, the access token has an expiration date. The duration of an access token, together with a refresh token, is provided when the user retrieves a new access token.

This PR aims to:

* properly store the information about the expiration of a retrieved access token
* expose, in the OAuth 2 client, a convenience mechanism to automatically renew an expired token by using the `refresh token`.
* improve error handling by using a convenience enum struct to store all error codes

# Why?

I'm working on a [library](https://www.github.com/fabiomassimo/GoogleAuthenticator) that helps developers to easily access Google services. I found in OAuthSwift a perfect fit for implementing the OAuth 2 process but I wanted to extend its capability with a solid mechanism of access token automatic renewal only based on the access token expiration.

# How would I use it?

To kick off a request that needs to be authorized simply call, on a OAuth 2 client, the following method:

```swift
oauthClient = OAuth2Swift(consumerKey: consumerKey, consumerSecret: consumerSecret, authorizeUrl: AuthenticationConstants.AuthorizeUrl.rawValue, accessTokenUrl: AuthenticationConstants.AccessTokensUrl.rawValue, responseType: AuthenticationConstants.ResponseType.rawValue)

oauthClient.startAuthorizedRequest(urlString, method: .GET, parameters: parameters, headers: headers, onTokenRenewal: { (credential) in
           // Convenience fallback to store new credential somewhere safe (i.e. the keychain)
            do {
                try self.credential.storeOAuthCredential(credential)
            } catch {
                failure(error: error as NSError)
            }
        }, success: success, failure: failure)
```

PR #596, opened 2020-05-12
Automatically refresh access tokens in OAuthSwiftClient

Timac:
# Summary

Access tokens have an expiration date but OAuthSwift doesn't automatically refresh access tokens by default.

This PR proposes a solution:
- that makes OAuthSwift automatically refresh access tokens when they expire
- which is opt-in by providing one simple convenient method
- containing the minimum amount of changes needed to avoid to break existing apps

Related links:
- Issue #217 
- PR #209 

Changes:
- Add a new function `requestWithAutomaticAccessTokenRenewal` in OAuthSwiftClient. When used, it automatically refreshes the access token if it expired and restart the query.
- Move the implementation for `renewAccessToken` and `requestOAuthAccessToken` from OAuth2Swift to OAuthSwiftClient. The methods `renewAccessToken` and `requestOAuthAccessToken` are still available in OAuth2Swift but they simply forward the call to OAuthSwiftClient to ensure that old apps don't need to be changed.

# Why?

I developed an application, called [Clatters](https://apps.apple.com/app/clatters/id1480930237), that needs to access various services relying on OAuth. Rather that creating my own OAuth library, I decided to use OAuthSwift as it perfectly fit my needs.

However I believe that OAuthSwift should automatically refresh access tokens when they expire. The app itself shouldn't have to implement the logic to handle expired token.

[Clatters](https://apps.apple.com/app/clatters/id1480930237) has been available in the iOS App Store since February 2020 and relies on the code written in this PR.

# Usage

You can call the convenient method `requestWithAutomaticAccessTokenRenewal` in OAuthSwiftClient with the correct parameters. If the access token expired, `requestWithAutomaticAccessTokenRenewal` will automatically renew the access token and retry the query. Here is an example to perform a query on Reddit: 

```
static func requestIdentity(client: OAuthSwiftClient) {
		client.requestWithAutomaticAccessTokenRenewal(url: URL(string: "https://oauth.reddit.com/api/v1/me")!, method: .GET, headers: nil, contentType: nil, accessTokenBasicAuthentification: true, accessTokenUrl: "https://www.reddit.com/api/v1/access_token", onTokenRenewal: nil) { (result) in
			switch result {
				case .success(let response):
					// Do something
					debugPrint("Received \(response)")
					break
				case .failure:
					// Fail nicely
					break
			}
		}
    }
```

# Notes

I wanted to keep the changes proposed in this PR really simple. As previously mentioned:

- this new feature is opt-in by using a convenient method.
- apps already using OAuthSwift shouldn't be affected and don't need to adopt the new API.
- no code has been removed or deprecated.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -72,14 +72,14 @@ open class OAuth2Swift: OAuthSwift {
     // MARK: functions
     @discardableResult
     open func authorize(withCallbackURL callbackURL: URLConvertible?, scope: String, state: String, parameters: Parameters = [:], headers: OAuthSwift.Headers? = nil, completionHandler completion: @escaping TokenCompletionHandler) -> OAuthSwiftRequestHandle? {
-      
+
         OAuthSwift.log?.trace("Start authorization ...")
         if let url = callbackURL, url.url == nil {
             completion(.failure(.encodingError(urlString: url.string)))
             return nil
         }
         self.observeCallback { [weak self] url in
-         
+
             OAuthSwift.log?.trace("Open application resource url: \(url.absoluteString)")
             guard let this = self else {
                 OAuthSwift.retainError(completion)
@@ -176,7 +176,7 @@ open class OAuth2Swift: OAuthSwift {
     }
 
     open func postOAuthAccessTokenWithRequestToken(byCode code: String, callbackURL: URL?, headers: OAuthSwift.Headers? = nil, completionHandler completion: @escaping TokenCompletionHandler) -> OAuthSwiftRequestHandle? {
-      
+
         var parameters = OAuthSwift.Parameters()
         parameters["client_id"] = self.consumerKey
         parameters["code"] = code
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -200,83 +200,11 @@ open class OAuth2Swift: OAuthSwift {
 
     @discardableResult
     open func renewAccessToken(withRefreshToken refreshToken: String, parameters: OAuthSwift.Parameters? = nil, headers: OAuthSwift.Headers? = nil, completionHandler completion: @escaping TokenCompletionHandler) -> OAuthSwiftRequestHandle? {
-        var parameters = parameters ?? OAuthSwift.Parameters()
-        parameters["client_id"] = self.consumerKey
-        parameters["client_secret"] = self.consumerSecret
-        parameters["refresh_token"] = refreshToken
-        parameters["grant_type"] = "refresh_token"
-        OAuthSwift.log?.trace("Renew access token, parameters: \(parameters)")
-        return requestOAuthAccessToken(withParameters: parameters, headers: headers, completionHandler: completion)
+        return self.client.renewAccessToken(accessTokenUrl: self.accessTokenUrl, withRefreshToken: refreshToken, parameters: parameters ?? OAuthSwift.Parameters(), headers: headers, completionHandler: completion)
     }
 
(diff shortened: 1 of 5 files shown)


phimage:
Thanks I will look at that closely

I see some issue with code formatting

Timac:
@phimage Let me know if I can help in any way or if you need some clarifications.

phimage:
I need that you fix the indentation/fomatting, I make a review about this
then I have time I will check the code closely (maybe this weekend)

Timac:
@phimage I pushed a first change to fix the swiftlint warnings introduced in https://github.com/OAuthSwift/OAuthSwift/commit/94f98f745f8d00824174b0d289f683c7043b2d2d

That's a change I didn't want to include in this PR, but I think it will help for you to review.

phimage:
thanks, I will check code format later

dutennakoon:
Hi, when can we expect this to be released?

Timac:
> Hi, when can we expect this to be released?

This PR has been merged so you can easily adopt the code today by pulling the code from the `master` branch.

dutennakoon:
Thanks @Timac. I was using the release number as the pod OAuthSwift version. `pod 'OAuthSwift', '~> 2.0.0'`

dutennakoon:
@Timac When can we expect a release tag for this? As we cannot control the versions from pulling the code from the `master` branch always.

phimage:
@dutennakoon 
You could specify a specific commit hash instead of master head (yes its better for security reason)

I use SwiftPM now (and sometimes carthage) so I will not be fast with cocoadpod (no more installed and connected to their specs)
But because of Xcode 12 I want to make a new release, publish older tag too
I have just loose all my id/token to test many services, it take time to recreate it....

dutennakoon:
Thanks for the update @phimage

vijaywargiya:
Hey @Timac, in your example you are passing onTokenRenewal as nil. But in the method you have the following code on line number 289. So, the renewal does not happen if you pass this as nil. Can you let me know if I'm doing something wrong here?

` if let onTokenRenewal = onTokenRenewal {`

vijaywargiya:
I have made the changes here and now this works correctly. https://github.com/vijaywargiya/OAuthSwift/commit/67b441aec715432996a871ee51ef13e76083dcbd

Timac:
@vijaywargiya The example could indeed be improved to demonstrate how to pass a `onTokenRenewal`.
The idea is that you have to pass a non-nil `onTokenRenewal` if you want the renewal to happen. If you pass nil, the renewal should not happen.

The change you proposed could led to more issues than it helps. I believe that after this change, `requestWithAutomaticAccessTokenRenewal` could potentially call itself with the same parameters over and over. This would led to an infinite loop and the app to crash.

I would suggest to instead make sure that you pass a non-nil `onTokenRenewal` if you want to renewal feature to happen.

vijaywargiya:
I think in that case we should make the parameter not accept nil. Since it would not work with a nil param anyways.
Also, I suppose this `onTokenRenewal` would not be useful everytime and we should accept and handle nil `onTokenRenewal`. We can add handling so that `requestWithAutomaticAccessTokenRenewal` is not called in a loop again and again.

gadget-man:
I'm looking for exactly this functionality, but can't work out how to use it!! Currently I call `oauthswift.authorize` which returns the access_token, refresh_token and expiry, which I save to the keychain. What I can't work out is how to replace oauthswift.authorize with requestWithAutomactiAccesstokenRenewal - can somebody point me in the right direction?

phatblat:
This introduced a new API and thus warrants a new minor version. Going to change the version from 2.1.3 to 2.2.0.

PR #605, opened 2020-06-21
Fix broken build when compiling for Mac Catalyst

Timac:
The revision `42e18ec60f45e1f51585410cd63758c1b47574cc` introduced a build error when compiling for Mac Catalyst.

As you can see in `SFAuthenticationURLHandler.swift`, SFAuthenticationURLHandler is not defined on Catalyst:

```
#if !targetEnvironment(macCatalyst)
@available(iOS, introduced: 11.0, deprecated: 12.0)
open class SFAuthenticationURLHandler: OAuthSwiftURLHandlerType {
[...]
#endif
```

The proposed solution consists of using the same `#if targetEnvironment(macCatalyst)` to not use SFAuthenticationURLHandler when compiling for Catalyst.

### DIFF
--- a/Sources/OAuth2Swift.swift
+++ b/Sources/OAuth2Swift.swift
@@ -137,6 +137,14 @@ open class OAuth2Swift: OAuthSwift {
                     let codeString = responseParameters["error_code"],
                     let code = Int(codeString) {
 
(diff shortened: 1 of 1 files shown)


phimage:
Thanks

PR #612, opened 2020-07-22
Fix for compiling Mac Framework

tevendale:
Unable to compile the Mac Framework Target in the latest build as per issue #661

I've popped in some platform-specific macros to remove the failing lines fro the Mac build, but I'm not sure if there are further implications from this change.

Also, OAuthLogProtocol.swift has been included in the Mac target for the framework


phimage:
thanks

Issue #613, opened 2020-07-25
Calling Convenience initialiser when subclassing OAuth2Swift fails from xcode 11.4 onwards


jlynch1:
### Description:

**Problem:**
Calling convenience initialisers when subclassing `OAuth2Swift` fails from xcode 11.4 onwards. In my case I was using the convenience init with parameter `accessTokenUrl` which was fine prior to xcode 11.4

https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes
Convenience initializer inheritance for subclasses defined outside the module that defines the base class now comes with additional restrictions. When these subclasses have a base class with non-public designated initializers, they no longer automatically inherit convenience initializers from their superclasses. To restore this automatic inheritance behavior, the base class must ensure that all of its designated initializers are public or open. (51249311)

**Fix/Workaround for 2.0 and 2.1 branch**
As `OAuth2Swift` is a subclass of `OAuthSwift`, the init() of `OAuthSwift` , the base class, needs to be made public. That way I can call the convenience initialisers on a subclass of `OAuth2Swift` from xcode 11.4 onwards. 

The version on master works around this by having just a single initialiser (with no convenience initialisers) in `OAuth2Swift` where `accessTokenUrl` is an optional default parameter with default value of nil. 

**Aside:**
Reason why I need to subclass is due to an issue when calling the method `renewAccessToken()` of class `OAuth2Swift` externally from my Application. It is the same as https://github.com/OAuthSwift/OAuthSwiftAlamofire/issues/11 and the workaround is to override `renewAccessToken()` to set `client.credential.oauthToken = ""`

### OAuth Version:
- [ ] Version 1
- [x] Version 2

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [x] CocoaPods
- [ ] Swift Package Manager
- [ ] Manually

### Library version:
- [ ] head
- [x] v2.1.0
- [x] v2.0.0
- [ ] v1.4.1
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 11.4 (Swift 5.2)
- [ ] 11.x (Swift 5.1)
- [ ] 10.x (Swift 5.0)
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

jlynch1:
[...]

PR #617, opened 2020-07-27
#613 Calling Convenience initialiser when subclassing OAuth2Swift fai…

jlynch1:
…ls from xcode 11.4 onwards


phatblat:
Fixes #613

PR #623, opened 2020-09-07
Fix the signature format for PLAINTEXT

chrishannah:
I encountered an issue regarding the signature format when using PLAINTEXT as the signature method. And after doing some research and debugging, I found that OAuthSwift was formatting the signature incorrectly. 

While this may not be the best solution, I have tested it with the (private) api that uses PLAINTEXT, and that does fix the issue.

Just for reference, the signature format for plaintext is highlighted in the [Dropbox documentation](https://dropbox.tech/developers/using-oauth-1-0-with-the-plaintext-signature-method).


phimage:
I think I never implemented it
https://github.com/OAuthSwift/OAuthSwift/blob/77a861bb6bd12a806451301a44d238e65ec568bb/Sources/OAuthSwiftCredential.swift#L110
 just make the enum to introduce other method like rsa

thanks

Issue #615, opened 2020-07-26
Release version 2.1.2 to CocoaPods


moberwasserlechner:
### Description:

Would it be possible to release the latest bugfix release to *CocoaPods.org*?

Currently only version 2.1.0 can be obtained from there. (see https://cocoapods.org/pods/OAuthSwift) 

Thx & BR, Michael

### OAuth Version:
- [ ] Version 1
- [x] Version 2

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [x] CocoaPods
- [ ] Swift Package Manager
- [ ] Manually

phimage:
I will do when I have some time, If I can I prefer to make some test before releasing. 
You could use HEAD code in podfile

ghowen:
@phimage Nice one! I always seem to forget this. Just on a side note: `:head` is deprecated and has been replaced with `:git`. So to get the master branch one would add

`  pod 'OAuthSwift', :git => 'https://github.com/OAuthSwift/OAuthSwift.git'
`

to one's `podfile`. 

Anyways, thanks for the library. OAuth1ing to OpenStreetMap is a breeze with this 👍

moberwasserlechner:
@ghowen @phimage Thank you both I didn't know that this was possible with cocoapods.

vgritsenko:
While it is possible to use `:git => ... branch => ` with CocoaPods in the Podfile, the same is not possible for other pods which declare OAuthSwift as a dependency. As a result, any such pod would fail validation and it would not be possible to release such pod automatically:

```
pod repo push <repo> MyAwesomePod.podspec --<parameters>
```

In our case we are waiting for the 2.1.3 release which brings in addition for token refresh. Would it be possible to release 2.1.3 sooner?

https://github.com/OAuthSwift/OAuthSwift/pull/620

macbellingrath:
@phimage any update on the release?

macbellingrath:
@phimage would a release be considered for sponsors?

phatblat:
@macbellingrath 2.1.2 was release a couple days ago

PR #643, opened 2021-01-19
Restore lint-ability to OAuthSwift Podspec

macbellingrath:
Background:
The podspec definition on the main branch lists support for iOS, watchOS, tvOS, macOS  but if one creates a target (as CocoaPods does during lint) for watchOS or tvOS, it can be seen that ASWebAuthenticationURLHandler isn't supported on those platforms. 

Because of how the conditional compilation was written - (nested `#if `) the entire class is actually only available on iOS currently. It would appear that the dependency ASWebAuthenticationURLHandler has on ASWebAuthenticationSession is available for macCatalyst targets, however, and I've left that intact. 

The Podspec is also failing to list its system framework dependencies. For iOS 9..<12 targets,  OAuthSwift depends on SafariServices, and on 12+, it also depends on AuthenticationServices. I've added those to the Podspec. 

With these changes - the pod can be pushed to specs and will succeed to lint. 

Thank you for contributing to OSS and taking time to review this diff. 

Related issues: 
https://github.com/OAuthSwift/OAuthSwift/issues/615

### DIFF
--- a/OAuthSwift.podspec
+++ b/OAuthSwift.podspec
@@ -9,6 +9,9 @@ Pod::Spec.new do |s|
     'Dongri Jin' => 'dongrify@gmail.com',
     'Eric Marchand' => 'eric.marchand.n7@gmail.com'
   }
+  s.weak_frameworks = 'AuthenticationServices'
+  s.ios.frameworks = 'SafariServices'
+  s.osx.frameworks = 'SafariServices'
   s.source = { git: 'https://github.com/OAuthSwift/OAuthSwift.git', tag: s.version }
   s.source_files = 'Sources/**/*.swift'
   s.swift_versions = ['5.0', '5.1']
--- a/Sources/Handler/ASWebAuthenticationURLHandler.swift
+++ b/Sources/Handler/ASWebAuthenticationURLHandler.swift
@@ -6,14 +6,13 @@
 //  Copyright © 2019 Dongri Jin, Marchand Eric. All rights reserved.
 //
 
-import Foundation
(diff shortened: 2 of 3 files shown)


phatblat:
Thanks for the PR! I'm going to be looking close at the podspec as I release the last couple tags to CocoaPods. Looks like I'll need this to get them published to trunk!

phatblat (approved):

phatblat:
This certainly solves the errors, but one warning (unrelated to this PR) remains.

```
↪ pod lib lint
    ...
    - WARN  | xcodebuild:  /Users/phatblat/dev/swift/OAuthSwift/Sources/Handler/OAuthWebViewController.swift:23:49: warning: using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead
    ...
    - ERROR | xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - ERROR | xcodebuild:  /Users/phatblat/dev/swift/OAuthSwift/Sources/OAuth2Swift.swift:153:25: error: cannot find 'ASWebAuthenticationURLHandler' in scope
    - ERROR | xcodebuild:  /Users/phatblat/dev/swift/OAuthSwift/Sources/OAuth2Swift.swift:156:25: error: cannot find 'SFAuthenticationURLHandler' in scope

[!] OAuthSwift did not pass validation, due to 3 errors and 1 warning.
```

phatblat:
The warning is new in Xcode 12.5; it does not appear when building with 12.4.

Update: warning fixed in #650

PR #647, opened 2021-03-02
Code verifier & code challenge generator for working with PKCE

MrWoWander:
To work with PKCE, it is more convenient to have a code verifier and code challenge generator in the framework. 
Code challenge generation works according to the "SHA256" standard

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,8 @@ let handle = oauthswift.authorize(
 ```
 
 ### Authorize with OAuth2.0 and proof key flow (PKCE)
+
+*The code verifier generation and code challenge generation methods are ***not compatible*** with ```Linux```*
 ```swift
 // create an instance and retain it
 oauthswift = OAuth2Swift(
@@ -209,7 +211,6 @@ let handle = oauthswift.authorize(
 }
 
 ```
-
 
 See demo for more examples
 
--- a/README.md
+++ b/README.md
@@ -179,8 +179,6 @@ let handle = oauthswift.authorize(
 ```
 
 ### Authorize with OAuth2.0 and proof key flow (PKCE)
-
-*Methods ```generateCodeVerifier``` and ```generateCodeChallenge``` are ***not compatible*** with ```Linux```*
 ```swift
 // create an instance and retain it
 oauthswift = OAuth2Swift(
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ let handle = oauthswift.authorize(
 
 ### Authorize with OAuth2.0 and proof key flow (PKCE)
 
-*The code verifier generation and code challenge generation methods are ***not compatible*** with ```Linux```*
+*Methods ```generateCodeVerifier``` and ```generateCodeChallenge``` are ***not compatible*** with ```Linux```*
 ```swift
 // create an instance and retain it
 oauthswift = OAuth2Swift(
--- a/README.md
+++ b/README.md
@@ -189,8 +189,8 @@ oauthswift = OAuth2Swift(
 )
 oauthswift.accessTokenBasicAuthentification = true
 
-let codeVerifier = base64url("abcd...")
-let codeChallenge = codeChallenge(for: codeVerifier)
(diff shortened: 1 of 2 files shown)


phimage:
thanks, every body could copy you code if I do not merge it 
(similar I think than the code in auth0 doc)

The reason that I do not merge immediately is the grip on `CommonCrypto`
we could add maybe macro if canImport

even the lib I think is not linux compatible this add a new issue
for linux and swiftpm there is 
https://github.com/apple/swift-crypto
I do it one time here https://github.com/IBDecodable/IBLinter/blob/master/Package.swift#L59 , but maybe there is more clean way now because SwiftPM had some evolution

MrWoWander:
> thanks, every body could copy you code if I do not merge it
> (similar I think than the code in auth0 doc)
> 
> The reason that I do not merge immediately is the grip on `CommonCrypto`
> we could add maybe macro if canImport
> 
> even the lib I think is not linux compatible this add a new issue
> for linux and swiftpm there is
> https://github.com/apple/swift-crypto
> I do it one time here https://github.com/IBDecodable/IBLinter/blob/master/Package.swift#L59 , but maybe there is more clean way now because SwiftPM had some evolution

Then I have a suggestion to add a new library to SPM: Perfect-Crypto. She herself knows how to generate SHA256
https://github.com/PerfectlySoft/Perfect-Crypto#digest

Well, or I can try to get confused, and write my translation in SHA256 by analogy, as in this library =)

phimage:
same here, I do not want to get all crypto function for one need (common crypto is at least embedded in iOS)

If a person have a have with common crypto, crypto swift, or perfect-crypto already in its app I do not want add a new lib

MrWoWander:
Okay. I have limited the given methods from Linux.

> same here, I do not want to get all crypto function for one need (common crypto is at least embedded in iOS)
> 
> If a person have a have with common crypto, crypto swift, or perfect-crypto already in its app I do not want add a new lib
> 
> (I have same issue with RSA)

MrWoWander:
I suggest this as a temporary solution, as several people have already encountered ```PKCE```

phatblat:
This could be made portable to Linux if you use Swift [Crypto](https://swift.org/blog/crypto/) instead of CommonCrypto

MrWoWander:
> This could be made portable to Linux if you use Swift [Crypto](https://swift.org/blog/crypto/) instead of CommonCrypto

On your advice, it seems like it turned out to implement support for SHA256

phimage:
I think it's ok and could be merged

I am just wondering if we keep as global function or put at static extension of OAuthSwift class

PS: Code do not compile on linux yet, a lot of `if os(target)` , import of uikit and appkit
things like importing `FoundationNetworking` I think also

I have made a test here https://github.com/OAuthSwift/OAuthSwift/runs/2465840429 or here https://github.com/OAuthSwift/OAuthSwift/runs/2465733255

README.md
```diff
@@ -179,6 +179,8 @@ let handle = oauthswift.authorize(
 ```
 
 ### Authorize with OAuth2.0 and proof key flow (PKCE)
+
+*Methods ```generateCodeVerifier``` and ```generateCodeChallenge``` are ***not compatible*** with ```Linux```*
```
phimage, line 183: Could be removed now?

Sources/Utils.swift
```diff
@@ -21,3 +27,42 @@ public func generateState(withLength len: Int) -> String {
     }
     return randomString
 }
+
```
phimage, line 30: only one blank line, I think SwiftLint will warn about it

phimage (changes requested):

README.md
MrWoWander: Done

Sources/Utils.swift
MrWoWander: Done

phimage:
I merge it, I will fix other SwiftLint warnings using `swiftlint autocorrect`

PR #650, opened 2021-04-29
🚨 Fix warning

phatblat:
This is a new warning in Xcode 12.5.

> OAuthWebViewController.swift:23:49: Using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead


phimage:
thanks

I have tested with Xcode 12.4

ps: I will try to fix the ci (GitHub action and travis)

Issue #621, opened 2020-08-26
ASWebAuthenticationSession support


raxityo:
Hi, 

I see that we have [`ASWebAuthenticationURLHandler`](https://github.com/OAuthSwift/OAuthSwift/blob/master/Sources/Handler/ASWebAuthenticationURLHandler.swift) that nicely supports `ASWebAuthenticationSession`.
However, it has two limitations:
1. It's marked as `@available(iOS 13.0, ...)`.
	- It could work in `iOS 12.0+` by preventing the use of `presentationContextProvider` and `prefersEphemeralWebBrowserSession`.
2. It requires users to declare URI schemes in `Info.plist` and update the app delegate to hand-off the callback to `OAuthSwift`.
	- This can be prevented by directly handing off the callback in the completion handler, making the overall setup and usage of OAuthSwift much simpler.

To overcome these limitations, I have added a class [here](https://github.com/raxityo/OAuthSwiftAuthenticationServices/blob/1.0.0/Sources/OAuthSwiftAuthenticationServices/ASWebAuthenticationSessionURLHandler.swift). But if you guys think it can be helpful to be added in this library, I can create a PR in this repository.

Thanks!

jrtibbetts:
So was this class ever added to OAuthSwift? I wish it were.

phatblat:
@jrtibbetts [`ASWebAuthenticationURLHandler`](https://github.com/OAuthSwift/OAuthSwift/blob/master/Sources/Handler/ASWebAuthenticationURLHandler.swift) is the library's wrapper for `ASWebAuthenticationSession`. It has been a part of the library for a couple of years now.

phatblat:
@raxityo If we downgrade the minimum version of iOS that this handler requires then we can't expose those newer properties you mention. I for one am looking forward to using the `prefersEphemeralWebBrowserSession` property to get rid of the alert that appears before the web page is loaded.

Looking at your code, you didn't actually prevent the use of these properties, but rather only used them when on iOS 13.0+. This is a nice touch, but this OS-version-dependent behavior should certainly be documented. Would you be interested in opening a PR with this change?

Regarding the direct callback handoff, I am unsure whether this is the right way to handle this. When an app is using a custom URL scheme, this provides better security as it avoids the potential security issue where a second app implements the same URL scheme and intercepts the redirect.

However, if an app uses Universal Links (https://example.com/oauth2redirect?code=...), this might introduce an attack vector by not going through the iOS URL-loading system.

[...]

PR #654, opened 2021-05-03
🐛 Add colon after scheme in error redirect URL

### LINKED ISSUE #653: 🐛 Invalid URL built when ASWebAuthenticationSession returns an error
### Description:

When a non-nil error is returned to the `ASWebAuthenticationSession` completion handler provided by this library, a URL is built using only the scheme and some query string parameters.

```swift
let urlString = "\(self.callbackUrlScheme)?error=\(msg ?? "UNKNOWN")"
```

This results in a URL like the following being constructed:

```
"customScheme?error=The%20operation%20couldn%E2%80%99t%20be%20completed.%20(com.apple.AuthenticationServices.WebAuthenticationSession%20error%201.)"
```

Note that the `guard let url = URL(string: urlString)` is not sufficient to catch this because the `url` variable does end up with an instance of `URL`. The problem comes when the URL is opened with `UIApplication.shared.open(url)`:

> 2021-05-03 15:18:43.500540-0600 KPConsumerAuth_Example[65804:1976592] [default] Failed to open URL customScheme?error=The%20operation%20couldn%E2%80%99t%20be%20completed.%20(com.apple.AuthenticationServices ... or%201.): Error Domain=NSOSStatusErrorDomain Code=-50 "invalid input parameters" UserInfo={NSDebugDescription=invalid input parameters, _LSLine=234, _LSFunction=-[_LSDOpenClient openURL:options:completionHandler:]} 

Note that the success case is not affected as the `callback` URL passed to the completion handler is opened without any modification.

```swift
UIApplication.shared.open(successURL, options: [:], completionHandler: nil)
```

### OAuth Provider? (Twitter, Github, ..):
PingFederate

### OAuth Version:
- [ ] Version 1
- [x] Version 2

- [x] iOS : 14.5
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [x] CocoaPods
- [ ] Swift Package Manager
- [ ] Manually

### Library version:
- [ ] head
- [x] v2.1.0
- [ ] v2.0.0
- [ ] v1.4.1
- [ ] other: (Please fill in the version you are using.)

### Xcode version:
- [x] 12.5 (Swift 5.4)
- [ ] 11.4 (Swift 5.2)
- [ ] 11.x (Swift 5.1)
- [ ] 10.x (Swift 5.0)
- [ ] other: (Please fill in the version you are using.)

- [ ] objective c

phatblat:
Added a colon after the custom URL scheme given to `ASWebAuthenticationURLHandler`. This scheme is relayed to the [`ASWebAuthenticationSession` constructor](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/2990952-init), but is also used to build a redirect URL when there is an error returned to the `completionHandler`. While the resulting `urlString` creates a valid URL, this URL can't be opened by UIKit without this colon.

Fixes #653

### DIFF
--- a/Sources/Handler/ASWebAuthenticationURLHandler.swift
+++ b/Sources/Handler/ASWebAuthenticationURLHandler.swift
@@ -36,7 +36,7 @@ open class ASWebAuthenticationURLHandler: OAuthSwiftURLHandlerType {
                     let msg = error.localizedDescription.addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)
                     let errorDomain = (error as NSError).domain
                     let errorCode = (error as NSError).code
-                    let urlString = "\(self.callbackUrlScheme)?error=\(msg ?? "UNKNOWN")&error_domain=\(errorDomain)&error_code=\(errorCode)"
+                    let urlString = "\(self.callbackUrlScheme):?error=\(msg ?? "UNKNOWN")&error_domain=\(errorDomain)&error_code=\(errorCode)"
                     let url = URL(string: urlString)!
                     #if !OAUTH_APP_EXTENSIONS
                     UIApplication.shared.open(url, options: [:], completionHandler: nil)


phatblat:
Should [`SFAuthenticationURLHandler`](https://github.com/OAuthSwift/OAuthSwift/blob/e09698d5f9e106353544fa586edd408da869158d/.github/CONTRIBUTING.md) be updated as well? [`SFAuthenticationSession`](https://developer.apple.com/documentation/safariservices/sfauthenticationsession) was removed from iOS as of 13.

phatblat:
Build failed due to an issue with Danger:

> Failing the build, Danger cannot run without API access.
> You can see more information at https://danger.systems/guides/getting_started.html
> The command "bundle exec danger" exited with 1.

I see that `DANGER_GITHUB_API_TOKEN` is declared in [`.travis.yml`](https://github.com/OAuthSwift/OAuthSwift/blob/master/.travis.yml#L25), but it doesn't have a value. If this is an [encrypted variable](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml) then it sounds like those don't work for PRs anyway.

> Encrypted variables are not available to untrusted builds such as pull requests coming from another repository.

phimage:
`ASWebAuthenticationURLHandler` has been implemented by contributor, so I have not enough knowledge :(
So I will merge

But I read this one #621
I think we do not need to go out of the app to enter in again using the open url because we have the hand of the code with this completion handler, we could call `OAuthSwift.handle(url: url)`: (https://github.com/raxityo/OAuthSwiftAuthenticationServices/blob/1.0.0/Sources/OAuthSwiftAuthenticationServices/ASWebAuthenticationSessionURLHandler.swift#L33)

phatblat:
We are using `ASWebAuthenticationURLSession` and I have some questions out to Apple about how it is supposed to work with Universal Links, which we plan to switch to. I can help test and improve our handler.

PR #655, opened 2021-05-13
optionally change authorize url param names

daironmichel:
@phimage you were right to reject my original pull request. It's was dumb on my part.

Your solution is better and did help in part. However it didn't allowed me to change the name of the params in the authorize url.

The idea is that instead of having the standard param names:
`https://www.someprovider.com/authorize?oauth_token=<token>&oauth_consumer_key=<key>`
You can change them to:
`https://www.someprovider.com/authorize?token=<token>&key=<key>`
which is how E*Trade requires it to be.

Thanks for your help!


phimage:
merged!

PR #658, opened 2021-05-18
Fix Mac catalyst compilation

phimage:


phimage:
waiting for GitHub ci build before merging
there is some issue now, with job cancelled

PR #660, opened 2021-05-19
👷🏻‍♀️ Run Danger on GitHub Action

phatblat:
The Travis build is always blowing up with a "Bad credentials" error. 

> GET https://api.github.com/repos/OAuthSwift/OAuthSwift/pulls/659: 401 - Bad credentials // See: https://docs.github.com/rest (Octokit::Unauthorized)

Not sure how well this is going to work in a GitHub Action workflow, but with the automatic availability of the `GITHUB_TOKEN` for an action, it sure seems simpler to manage.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,0 +1,37 @@
+# https://github.com/marketplace/actions/danger-action
+# https://github.com/MeilCli/danger-action
+# https://github.com/MeilCli/danger-action/releases
+
+name: Danger
+
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  danger:
+    name: danger
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


phatblat:
The success from the new Danger check is misleading as there was an error running swiftlint.

> From https://github.com/OAuthSwift/OAuthSwift
>  * [new branch]      ci/action/danger -> origin/ci/action/danger
> /home/runner/work/OAuthSwift/OAuthSwift/vendor/bundle/ruby/2.6.0/gems/danger-swiftlint-0.26.0/ext/swiftlint/bin/swiftlint: 1: Syntax error: "(" unexpected

I didn't realize this was also going to run swiftlint, so will need to switch to a macOS runner.

Nef10:
Does danger still need to run SwiftLint, given that it already runs in its own action?

phimage:
I have never configured it, but I see danger SwiftLint, I have already launch SwiftLint on linux , I see some  docker and GitHub action about it

PR #665, opened 2021-06-11
Added custom log action

Vortec4800:
The new logging mechanism looks really useful, however, I wanted to extend it to integrate with my app's existing logging system. I added a custom log action that accepts a closure, and the log action is used instead of just printing to the console so the actual log can be customized.

This extends the work done by @svoip and the logging system.

In order to use this new functionality, you would add something like this:

```swift
OAuthSwift.setLogLevel(.trace)
OAuthSwift.setCustomLogAction { message in
   CustomLogger("[OAUTHSWIFT] \(message)")
}
```

### DIFF
--- a/Sources/OAuthLogProtocol.swift
+++ b/Sources/OAuthLogProtocol.swift
@@ -22,6 +22,8 @@ public enum OAuthLogLevel: Int {
 public protocol OAuthLogProtocol {
 
    var level: OAuthLogLevel { get }
+	
+   var logAction: ((_ message: String) -> Void)? { get set }
 
    /// basic level of print messages
    func trace<T>(_ message: @autoclosure () -> T, filename: String, line: Int, function: String)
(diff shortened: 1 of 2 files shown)


phimage:
Hi, 
I think it’s useless. OAuthLogProtocol is a protocol. Make your class or structure that implement it and put an instance in OAuthSwift.log

Vortec4800:
That's the path I started to go down before I did this. Unfortunately, it doesn't work for 2 reasons.

The first one, OAuthSwift.log isn't a public property, so you would need a PR on OAuthSwift to add a func to set a custom logging subclass anyway.

The second issue is the default implementations are in an extension on the protocol. If you create a subclass and set it as the logging class, Swift still calls the implementations in the extension and not those in the subclass. It might be possible to reorganize everything to make it work, but the few things I tried caused errors due to the default parameters to get file and line information.

This seemed like the best way to get custom logging functionality without totally re-writing the logger which didn't seem necessary.

Vortec4800:
Hi, any more info on this PR? I still think it's useful functionality to be able to customize logging, but the proposed solution of using the protocol isn't possible.

PR #672, opened 2021-08-12
Make TimeoutInterval configurable to fix issue OAuthSwift/issues/667

MikeSpelling:
Rather than providing a default timeout in the Config initialiser which is never overriden I've moved this up to the top level requests. This should mean its a non-breaking change but allows a timeoutInterval to be passed where needed and used by the UrlRequest.
This is my first open source contribution so I welcome any feedback - thanks!

### DIFF
--- a/Sources/OAuth1Swift.swift
+++ b/Sources/OAuth1Swift.swift
@@ -74,7 +74,7 @@ open class OAuth1Swift: OAuthSwift {
     // MARK: functions
     // 0. Start
     @discardableResult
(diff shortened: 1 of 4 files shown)


MikeSpelling:
@phimage any thoughts on this PR?

MikeSpelling:
Any update on this as yet?

vgritsenko:
A thought from a peanut gallery... This seems to add an extra parameter into a lot of methods. Wouldn't it be better to subclass `OAuthSwiftClient` and override open `makeRequest` function and tailor request to your needs in there? It would need a few small changes to `OAuthSwiftHTTPRequest`, but it would result in much less changes overall.

Issue #668, opened 2021-07-13
Feature Request - Add id_token support to OAuthSwiftCredential


MikeSpelling:
### Description:
In order to support single-sign out with a redirect endpoint you need access to the original `id_token` received during authentication (https://identityserver4.readthedocs.io/en/latest/endpoints/endsession.html#refendsession).
It seems like this should be easy enough to add to the `OAuthSwiftCredential` object as an optional String.
Around line 122 of OAuth2Swift for example (where the `access_token` is set) the `id_token` could also be added?

### OAuth Provider? (Twitter, Github, ..):
Custom - using our own Identity Server

### OAuth Version:
- [ ] Version 1
- [x] Version 2

### OS (Please fill the version) :
- [x] iOS :
- [ ] OSX :
- [ ] TVOS :
- [ ] WatchOS  :

### Installation method:
- [ ] Carthage
- [x] CocoaPods
- [ ] Swift Package Manager
- [ ] Manually

### Library version:
- [ ] head
- [ ] v2.1.0
- [ ] v2.0.0
- [ ] v1.4.1
- [x] other: (v2.2.0)

### Xcode version:
- [ ] 11.4 (Swift 5.2)
- [ ] 11.x (Swift 5.1)
- [ ] 10.x (Swift 5.0)
- [x] other: (Xcode 12.5.1, Swift 5.4)
- [ ] objective c

PR #679, opened 2021-10-07
Added parsing of id_token, if present in access token response

colinhumber:
Adds support for parsing of `id_token`, if present, when fetching an access token. Details on why this is useful is in https://github.com/OAuthSwift/OAuthSwift/issues/668, specifically to support Single Sign Out with a redirect URI.


phimage:
Hi,

Ok we just decode it if present and store it in memory into `OAuthSwiftCredential`

No need to store it ? in keychain for instance with other credential object informations

There is support of `Codable` and old `NSCoder` too

ps: I could merge anyway but could wait

MikeSpelling:
Is this still being looked at? I also need the id_token for single sign-out so would appreciate the merge!

MikeSpelling:
Any reason this is still unmerged?

PR #681, opened 2021-11-01
Respect basic auth and content type in renewAccessToken

01100010011001010110010101110000:
`renewAccessToken` does not currently respect `contentType` or `accessTokenBasicAuthentification`, causing certain refresh operations to fail because they lack the auth header. This fixes that by passing those parameters through in the same manner as `requestOAuthAccessToken`


bfahsel:
+1 for this PR.

Fitbit Oauth2 API requires this behaviour.

PR #697, opened 2022-04-29
Remove showing the network activity status, deprecated sind iOS 13

cmittendorf:
This change removes changing the network activity indicator as the necessary api has been marked as deprecated in iOS 13.

Besides the old solution was not safely being called on the main thread, the `OAUTH_APP_EXTENSIONS` build flag is of no use when being used as a swift package.


vgritsenko:
Apple has deprecated it since iOS 13. May be it is time to remove it from the library entirely?

```
@property(nonatomic,getter=isNetworkActivityIndicatorVisible) BOOL networkActivityIndicatorVisible API_UNAVAILABLE(tvos) API_DEPRECATED("Provide a custom network activity UI in your app if desired.", ios(2.0, 13.0));
```

cmittendorf:
Yes, the functionality should be removed completely. I totally missed that the api is deprecated.

PR #713, opened 2024-02-12
Improved error information

hactar:
I was recently debugging OAuth server side issues and the error message did not contain all infos I needed to send to the server team: detailed information about the request was missing. I have modified the error creation so that it includes this info as well.


phimage:
did you see that some swift errors contain the request?

https://github.com/OAuthSwift/OAuthSwift/blob/2dcf2380cad4b656a3ff03d25a2a765e9adb6551/Sources/OAuthSwiftError.swift#L33

could add something like that in OAuthSwiftError extension
```swift
public var request: URLRequest? {
        switch self {
        case .requestError(_, let request): return request
        case .authorizationPending(_, let request): return request
        case .slowDown(_, let request): return request
        case .accessDenied(_, let request): return request
        default: return nil
        }
    }
```

hactar:
Ah, no I did not notice this - I just generically caught "Error" and after Mirroring it, did not see any request in it so I assumed it wasn't there. Thanks, this makes my PR obsolete then.

